PR #67, opened 2017-10-12
go back buttons added

azayneeva:


react-backend/client/src/components/admin/homeAdmin.js
```diff
@@ -43,6 +43,7 @@ export class HomeAdmin extends Component {
             ))}
           </tbody>
         </table>
+        <Link to="/"><button className="Button">Back to the main page</button></Link><br />
```
rachaelcodes: perhaps we could change the styling?

PR #85, opened 2017-11-10
Fix scan bug

RogeredBacon:
Fixed typos, removed console logs, cleaned up code, became grateful of Zooeys expertise

### DIFF
--- a/react-backend/client/src/components/visitors/qrcode.js
+++ b/react-backend/client/src/components/visitors/qrcode.js
@@ -54,20 +54,19 @@ function instascan() {
         reject('ADD LISTENER FAILED');
       }
     });
-    Instascan.Camera
-      .getCameras()
+    Instascan.Camera.getCameras()
       .then(cameras => {
         if (cameras.length > 0) {
           scanner.start(cameras[0]);
         } else {
           console.error('No cameras found.');
-          console.log('ERROR HAPPEINGN AT getCameras');
+          console.log('ERROR HAPPENING AT getCameras');
 
-          throw new Error('ERROR HAPPEINGN AT getCameras');
+          throw new Error('ERROR HAPPENING AT getCameras');
         }
       })
       .catch(err => {
-        console.log('ERROR HAPPEINGN AT getCameras');
+        console.log('ERROR HAPPENING AT getCameras');
         console.log(err);
         throw err;
       });
--- a/react-backend/client/src/components/visitors/qrcode.js
+++ b/react-backend/client/src/components/visitors/qrcode.js
@@ -1,48 +1,78 @@
 /*global Instascan*/
 import React, { Component } from 'react';
 import { PurposeButton } from './purposeButton';
-import { withRouter} from 'react-router-dom';
-
+import { withRouter } from 'react-router-dom';
 
 function getUserFromQRScan(content) {
+  console.log('getUserFromQRScan: ', content);
   return fetch('/getUsername', {
-    method: "POST",
+    method: 'POST',
     headers: {
-      'Content-Type': 'application/json'
+      'Content-Type': 'application/json',
     },
-    body: JSON.stringify({user:content})
+    body: JSON.stringify({ user: content }),
   })
-  .then(res=>res.json())
-  .catch((err) => {
-    console.log("error from getUserFromQRScan:  ", err);
-    throw err
-  })
+    .then(res => res.json())
+    .catch(err => {
+      console.log('error from getUserFromQRScan:  ', err);
+      throw err;
+    });
 }
 
 function instascan() {
-  return new Promise((resolve, reject)=>{
+  return new Promise((resolve, reject) => {
     let scanner = new Instascan.Scanner({
       video: document.getElementById('preview'),
-      scanPeriod: 5
+      scanPeriod: 5,
     });
     scanner.addListener('scan', function(content) {
+      console.log(document.getElementById('preview'));
       if (document.getElementById('preview')) {
-        document.getElementById('preview').pause();
-      }
-      resolve(content)
-    });
-    Instascan.Camera.getCameras().then((cameras) => {
-      if (cameras.length > 0) {
-        scanner.start(cameras[0]);
-      } else {
-        console.error('No cameras found.');
-      }
-    }).catch((err) => {
-      console.error(err);
-    });
-  })
-}
+        console.log('scanner listener: ', content);
+        //previously you were calling .pause on the video element here, this
+        //paused the video element but didn't do anything with the scanner that
+        //you created above
+        //So (i think) whenever you ran this it created a new scanner somehow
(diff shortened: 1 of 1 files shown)


react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -53,58 +63,72 @@ export class QRCode extends Component {
...
+          this.handleVideo();
+          return getUserFromQRScan(content);
+        })
+        .then(user => {
+          this.setState({
+            username: user.fullname.fullname,
+            hash: user.fullname.hash,
+          });
+        })
+        .catch(error => {
```
ZooeyMiller: This error is going to be catching everything that could go wrong inside of the instascan functions (anything that rejects or throws in there is going to hit this catch) so I would suggest doing something more meaningful than just a `console.log`, perhaps route the user to `/visitor/qrerror`?

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -1,49 +1,59 @@
...
-    }).catch((err) => {
-      console.error(err);
     });
-  })
+    Instascan.Camera.getCameras()
+      .then(cameras => {
+        if (cameras.length > 0) {
+          scanner.start(cameras[0]);
+        } else {
+          throw new Error('ERROR HAPPENING AT getCameras');
```
ZooeyMiller: Since this and line 52 are inside of a Promise that's being returned, may be better to `reject(err)` than `throw err`.

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -53,58 +63,72 @@ export class QRCode extends Component {
...
-      body: JSON.stringify(visitInfo)
-    }).then(
-      this.props.history.push('/visitor/end')
-    );
+      method: 'POST',
+      body: JSON.stringify(visitInfo),
+    })
+      .then(this.props.history.push('/visitor/end'))
+      .catch(error => {
+        console.log('ERROR HAPPENING AT FETCH /postActivity', error);
```
ZooeyMiller: if something goes wrong posting the activity to the server, what do you want to do? Rollback the entire log in? Prompt the user to log in again? May be worth considering, and put that functionality here.

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -53,58 +63,72 @@ export class QRCode extends Component {
...
-    .then((activities)=> {
-      this.setState({activities})
-    })
+      .then(res => res.json())
+      .then(res => res.activities)
+      .then(activities => {
+        this.setState({ activities });
+      })
+      .catch(error => {
+        console.log('ERROR HAPPENING AT FETCH', error);
```
ZooeyMiller: if there is a problem fetching activities, it may be worth sending the user back to the start of the login flow, or sending them to `/visitor/qrerror`

react-backend/client/src/components/visitors/qrcode.js
azayneeva: @ZooeyMiller , thank you so much for such a comprehensive review! please, have a look at the changes: line 93 forwards to login page. What do you think is better: to add a new component similar to "qrerror" that says "Sorry, we cannot record your purpose of visit, please log in again" or just forward to login page without explaining anything?

react-backend/client/src/components/visitors/qrcode.js
azayneeva: @Zooey, thank you a lot! added on line 125

react-backend/client/src/components/visitors/qrcode.js
ZooeyMiller: I think you definitely need to explain that the login was unsuccessful, as far as I understand it's an important part of the login process that the activity is tracked right? So a login should be considered unsuccessful if an activity isn't tracked? 

Rather than creating a new component, you could try passing options to `qrerror` depending on where the error occurred? And it could display a different error message depending on where the error happened?

Whatever the specific implementation is, I think it's important that the user isn't just redirected to login with no information, they need to be told that a problem occurred.

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -94,6 +90,8 @@ export class QRCode extends Component {
       .then(this.props.history.push('/visitor/end'))
       .catch(error => {
         console.log('ERROR HAPPENING AT FETCH /postActivity', error);
+        this.props.history.push('/visitor/login');
+        throw error;
```
ZooeyMiller: If you are sending the user to another route (`/visitor/login/` or `/visitor/qrerror` or anything), you shouldn't also need to `throw` here, you are handling your error by redirecting the user, so the throw is unnecessary as far as I can see.

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -44,13 +44,9 @@ function instascan() {
         if (cameras.length > 0) {
           scanner.start(cameras[0]);
         } else {
-          throw new Error('ERROR HAPPENING AT getCameras');
+          reject('ERROR HAPPENING AT getCameras');
         }
       })
-      .catch(err => {
```
ZooeyMiller: You still need to have a `.catch` here, whenever a function returns a `Promise` it needs to have a `.then` and a `.catch`. Inside of the `.catch` here you could `reject` since this is all happening inside of a `Promise` you are writing. Then when you call your `instascan()` function, any failures inside of it will trigger the `.catch` you have on [line 111](https://github.com/FAC-11/DataPower/blob/b3eceabfdb304e1633a3711c884a93056be5aa26/react-backend/client/src/components/visitors/qrcode.js#L111)

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -1,49 +1,55 @@
...
     });
-  })
+    Instascan.Camera.getCameras()
+      .then(cameras => {
+        if (cameras.length > 0) {
+          scanner.start(cameras[0]);
+        } else {
+          reject('ERROR HAPPENING AT getCameras');
+        }
+      })
```
ZooeyMiller: You still need to have a `.catch` here, whenever a function returns a `Promise` it needs to have a `.then` and a `.catch`. Inside of the `.catch` here you could `reject` since this is all happening inside of a `Promise` you are writing. Then when you call your `instascan()` function, any failures inside of it will trigger the `.catch` you have on [line 111](https://github.com/FAC-11/DataPower/blob/b3eceabfdb304e1633a3711c884a93056be5aa26/react-backend/client/src/components/visitors/qrcode.js#L111)

ZooeyMiller:
In general, changes so far look good, just remember all of your `Promises` need to have a `.catch` with meaningful error handling, and remove the `console.log`s (in general console.logs can be useful for debugging but probably shouldn't be in your master branch/deployed)

Issue #16, opened 2017-09-29
As a user I can print out my QR code combined with a logo


rachaelcodes:
- [x] check with Sonja what printing options we might be able to offer
- [x] Is there a module for creating a document and sending it to the printer?

rachaelcodes:
In case this helps, cheap domestic printers look like they can print on A7 size (checking the manual for a printer I bought for £30 8 years ago)

rachaelcodes:
have talked it through with @RogeredBacon and @azayneeva - perhaps this is something for a future sprint, as it involves hardware purchasing decisions

RogeredBacon:
As a user I can print out my QR code combined with a logo.

- [ ] Automatically combining QR Code image, logo image and other info
- [ ] Printing image

RogeredBacon:
So I've found two modules, One that can create a pdf combining images, text, formatting etc :https://www.npmjs.com/package/pdfmake

and one that can print pdf's, without having to leave the app:http://printjs.crabbly.com/#documentation
It can also print things straight from the server without having to download them locally

PR #87, opened 2017-11-15
Reformatted the pdf and changed the QR Codes colour

RogeredBacon:
#16


RogeredBacon:
Sweet, we expected as much

react-backend/client/src/components/visitors/main.js
```diff
@@ -123,7 +123,7 @@ class Main extends Component {
             <h1>Please tell us about yourself</h1>
             <form className="Signup" onChange={this.handleChange} onSubmit={this.handleSubmit}>
               <Select question="Select Your Sex" option="sex" choices={['male', 'female', 'prefer not to say']}/>
-              <Select question="Year of Birth" option="year" choices={[1980,1981,1982,1983,1984,1985]}/>
+              <Select question="Year of Birth" option="year" choices={[ 1900, 1901, 1902, 1903, 1904, 1905, 1906, 1907, 1908, 1909, 1910, 1911, 1912, 1913, 1914, 1915, 1916, 1917, 1918, 1919, 1920, 1921, 1922, 1923, 1924, 1925, 1926, 1927, 1928, 1929, 1930, 1931, 1932, 1933, 1934, 1935, 1936, 1937, 1938, 1939, 1940, 1941, 1942, 1943, 1944, 1945, 1946, 1947, 1948, 1949, 1950, 1951, 1952, 1953, 1954, 1955, 1956, 1957, 1958, 1959, 1960, 1961, 1962, 1963, 1964, 1965, 1966, 1967, 1968, 1969, 1970, 1971, 1972, 1973, 1974, 1975, 1976, 1977, 1978, 1979, 1980, 1981, 1982, 1983, 1984, 1985, 1986, 1987, 1988, 1989, 1990, 1991, 1992, 1993, 1994, 1995, 1996, 1997, 1998, 1999, 2000, 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019]}/>
```
ZooeyMiller: I'd recommend you use a function which returns an array of the years from 1900 to the current year, rather than hardcoding it (sure, you go up to 2019 here, but that only future proof you so far, and isn't very automated/programmatic). You could use a simple for loop in combination with `(new Date).getFullYear()` to get the current year. You could also use a generator function or recursion, if you wanted to take a different approach. 

[Here's an example of a sequence generator with recursion](https://repl.it/@zooeyxmiller/sequence-generator-example). 

I also looked into an example with generator functions, but I don't know them so well so would err on the side of caution if you decide to use them! 

This would also be a great, simple pure function to write a test for.

RogeredBacon: Thanks that's much neater and easier to read.

Cheers @oliverjam also!

oliverjam: Here's [some more explanation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) on why that works with `Array.from`

Basically the first argument can be any iterable, and if it has a length property that's how long the array will be. The second argument can be a map function to determine what each thing in the array is 👍

PR #88, opened 2017-11-16
Combined neat branch with printfix

RogeredBacon:
1. Print works
2. Error messages much neater
3. Implemented generator function for years
4. changed name of pdf file in email
5. CB email is now a env var


RogeredBacon:
Will do in the future. You'll be pleased to learn we intend to write lots of tests tomorrow, hopefully you're job will be a little easier

Issue #91, opened 2017-11-29
Fix and refactor error messages on signup page


RogeredBacon:
They're currently not working as hoped and the code is a bit of a mess and needs to be refactored from the tons of 'if' statements it is currently.

We need to work out exactly how we want multiple errors to be displayed:

1. What errors takes priority?
2. Should we check the password as we type?
3. Should we display multiple errors or one at a time?

ZooeyMiller:
some notes of things to refactor: 

https://github.com/FAC-11/DataPower/pull/92#discussion_r153777068

https://github.com/FAC-11/DataPower/pull/92#discussion_r154140436

https://github.com/FAC-11/DataPower/pull/92#discussion_r154141040

azayneeva:
https://egghead.io/lessons/egghead-make-dynamic-forms-with-react  - an example of how to display error messages dynamically

RogeredBacon:
Completed, opening new issue to implement validation as you type.

Issue #77, opened 2017-11-07
As a CB I can login to the app and be given the choice between visitor signup/login and the admin dashboard.


RogeredBacon:
As a CB I can login to the app and be given the choice between visitor signup/login and the admin dashboard.

Issue #76, opened 2017-11-07
As a CB I can securely register with the app, specifying the type of business.


RogeredBacon:
As a CB I can securely register with the app, specifying the type of business.

azayneeva:
![screen shot 2017-11-17 at 15 35 38](https://user-images.githubusercontent.com/24795752/33136783-fcd4cb48-cf9d-11e7-8431-593ebcf3e7a8.png)

PR #92, opened 2017-11-29
CB login & signup

azayneeva:
relates to issues #76, #77, #91 

The code is super messy, but it's functional :) We will go back and refactor it next week.

### DIFF
--- a/react-backend/client/src/components/authentication/logincb.js
+++ b/react-backend/client/src/components/authentication/logincb.js
@@ -1,0 +1,86 @@
+import React, { Component } from 'react';
+import { Input } from '../visitors/input';
+import { Button } from '../visitors/button';
+
+class CBlogin extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      email: '',
+      password: '',
+      error: [],
+    };
+  }
+
+  handleChange = e => {
+    let newState = {};
+    newState[e.target.name] = e.target.value;
+    this.setState(newState);
+  };
+
+  setError(messagesArray) {
+    this.setState({ error: messagesArray });
+  }
+
+  handleSubmit = e => {
+    e.preventDefault();
+
+    const checkData = {
+      formEmail: this.state.email,
+      formPswd: this.state.password,
+    };
+
+    fetch('/checkCBlogin', {
+      method: 'POST',
+      body: JSON.stringify(checkData),
+    })
+      .then(res => res.text())
+      .then(data => {
+        const EMAIL_ERROR = (
+          <span>
+            This email is invalid - please make sure you have entered a valid email address.
+          </span>
+        );
+        const DETAILS_ERROR = (
+          <span>The email address or password is incorrect. Please try again.</span>
+        );
+        const NO_INPUT_ERROR = (
+          <span>
+            Oops, you need to enter your information. <br />Please make sure you leave no input
+            field blank before continuing.
+          </span>
+        );
+        if (data === 'true') {
+          console.log("Success, now it's time to use cookies");
+          this.props.history.push('/visitor');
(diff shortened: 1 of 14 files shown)


react-backend/client/src/components/authentication/logincb.js
```diff
@@ -0,0 +1,86 @@
...
+        if (data === 'true') {
+          console.log("Success, now it's time to use cookies");
+          this.props.history.push('/visitor');
+        } else if (data === 'email') {
+          this.setError([EMAIL_ERROR]);
+        } else if (data === 'noinput') {
+          this.setError([NO_INPUT_ERROR]);
+        } else if (data === 'false') {
+          this.setError([DETAILS_ERROR]);
+        }
```
ZooeyMiller: this could be refactored into a case statement

react-backend/routes/cbauthentication/checkCB.js
```diff
@@ -0,0 +1,61 @@
...
+      data.formEmail.length === 0 ||
+      data.formGenre.length === 0 ||
+      data.formPswd.length === 0 ||
+      data.formPswdConfirm.length === 0
+    ) {
+      res.send('noinput');
+    } else if (validator.isEmail(data.formEmail) && validator.isAlpha(org_name, ['en-GB'])) {
+      getCBAlreadyExists(data.formName.toLowerCase(), data.formEmail, (error, result) => {
+        if (error) {
+          console.log('error from getCBAlreadyExists ', error);
```
ZooeyMiller: Add some meaningful error handling here. even just:
```js
res.status(500).send({error: errorStuff})
```
as if this errors, the browser will just hang with no response as it is at the moment.

react-backend/routes/cbauthentication/registerCB.js
```diff
@@ -0,0 +1,41 @@
+const express = require('express');
+
+const router = express.Router();
+
+const hashCB = require('../../functions/cbhash');
+const putCBData = require('../../database/queries/CBqueries/putCBData');
+const sendCBemail = require('../../functions/sendCBemail');
+
+let details = {};
```
ZooeyMiller: why the global?
does it not work if you set `const details = JSON.parse(body)` on line 21?

react-backend/client/src/components/authentication/logincb.js
```diff
@@ -0,0 +1,86 @@
...
+      formPswd: this.state.password,
+    };
+
+    fetch('/checkCBlogin', {
+      method: 'POST',
+      body: JSON.stringify(checkData),
+    })
+      .then(res => res.text())
+      .then(data => {
+        const EMAIL_ERROR = (
```
ZooeyMiller: since you use the same error info form `EMAIL_ERROR` here and in `signupcb` (and some other errors I think) it could be worth making something like an `errors.js` file and exporting all of them from there so that they are reusable.

react-backend/routes/cbauthentication/checkCBlogin.js
```diff
@@ -0,0 +1,36 @@
...
+    } else if (validator.isEmail(data.formEmail)) {
+      data.formPswd = hashCB(data.formPswd);
+      getCBlogindetailsvalid(data.formEmail, data.formPswd, (error, result) => {
+        if (error) {
+          console.log('error from getCBlogindetailsvalid ', error);
+        } else {
+          res.send(result.rows[0].exists);
+        }
+      });
+    } else if (!validator.isEmail(data.formEmail)) {
```
ZooeyMiller: this can just be an else as you will only ever hit it if `validator.isEmail(data.formEmail)` is falsey

azayneeva:
@ZooeyMiller thank you for so quick and helpful review !

Issue #17, opened 2017-09-29
As a user I can see an error page if there is a problem


rachaelcodes:
- [x] 404 (probably not likely, as the user should not be typing in different endpoints)
- [x] 500
- [x] connect these to database errors in database functions

azayneeva:
Done

PR #93, opened 2017-12-04
Changed the scema to add timestamps to all tables. Made email a uniqu…

RogeredBacon:
…e column of cbusiness table. Changed checkCB and SignupCB to only check if the email is already registered, not the email and org_name and give a relevant error message

#76
#17


react-backend/database/dbBuild.sql
```diff
@@ -1,67 +1,73 @@
...
   hash VARCHAR(64) NOT NULL
 );
-
-CREATE TABLE cbusiness (
+CREATE TABLE cbusiness
+(
   id SERIAL PRIMARY KEY,
   org_name VARCHAR(100) NOT NULL,
   genre VARCHAR(50) NOT NULL,
   email VARCHAR(100) NOT NULL,
```
ZooeyMiller: to make email unique you can change this line to:
```SQL
email VARCHAR(100) NOT NULL UNIQUE,
```
I assume the current method works too but I think this one is a little clearer?

react-backend/database/dbBuild.sql
```diff
@@ -1,67 +1,73 @@
...
-  hash_pwd VARCHAR(64) NOT NULL
+  hash_pwd VARCHAR(64) NOT NULL,
+  date TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  UNIQUE (email)
```
ZooeyMiller: to make email unique you can change this line to:
```SQL
email VARCHAR(100) NOT NULL UNIQUE,
```
I assume the current method works too but I think this one is a little clearer?

RogeredBacon:
Cool, changed it. It's definitely neater

PR #97, opened 2017-12-05
Newauth: backend and frontend routes are protected

azayneeva:
relates to issue #77


ZooeyMiller:
If a PR isn't ready for review/merge it's good practice to put `(WIP)` at the beginning of the title, and putting **not ready for review** in bold, at the top of the PR body to make it clear.

When it is ready, add an `awaiting-review` label, and remove the `(WIP)` and **not ready for review**.

azayneeva:
thank you for the comment, @ZooeyMiller! will follow this practice! :)

react-backend/app.js
```diff
@@ -32,7 +32,7 @@ app.use(cookieParser());
 app.use(express.static(path.join(__dirname, 'client/build')));
 
 app.use('/', index);
-app.use('/users', users);
+// app.use('/users', users);
```
ZooeyMiller: is this meant to be commented out?

react-backend/client/src/components/authentication/logincb.js
```diff
@@ -51,14 +51,15 @@ class CBlogin extends Component {
...
+          this.props.setLoggedIn();
           this.props.history.push('/visitor');
-        } else if (data === 'email') {
+        } else if (data.reason === 'email') {
           this.setError([EMAIL_ERROR]);
-        } else if (data === 'noinput') {
+        } else if (data.reason === 'noinput') {
           this.setError([NO_INPUT_ERROR]);
-        } else if (data === 'false') {
+        } else if (data.success === false) {
```
ZooeyMiller: this can just be an else statement. 

The only way the code would ever get to this condition is if `data.success` isn't true (since it's always a boolean, that means it must be false), and if `data.reason` is not 'email' or 'noinput', so you don't have to check that `data.success === false` here, as if it gets to this bit of code, it always will be false.

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -86,8 +88,13 @@ export class QRCode extends Component {
       activity: newActivity,
     };
 
+    const headers = new Headers({
```
ZooeyMiller: rather than rewrite this bit of code twice (here and line 126), you could put the headers in the class (rather than each method in the class), and then use them as `this.headers` in the fetch request. 

Would just mean a little less repeated code.

react-backend/routes/getAllUsers.js
```diff
@@ -1,15 +1,24 @@
...
+  jwt.verify(req.headers.authorization, process.env.SECRET, (err, payload) => {
+    if (err) {
       console.log(err);
-    });
+      res.send(JSON.stringify({ error: 'Not logged in' }));
+    } else {
+      getAllUsers()
+        .then(users => res.send({ users }))
+        .catch((err) => {
+          console.log(err);
```
ZooeyMiller: in this error case we need to send a 500 error to the front end.

react-backend/routes/activities.js
```diff
@@ -1,15 +1,24 @@
...
+  jwt.verify(req.headers.authorization, process.env.SECRET, (err, payload) => {
+    if (err) {
       console.log(err);
-    });
+      res.send(JSON.stringify({ error: 'Not logged in' }));
+    } else {
+      activities()
+        .then(activities => res.send({ activities }))
+        .catch((err) => {
+          console.log(err);
```
ZooeyMiller: as mentioned in `getAllUsers.js` we need to send a 500 error to the front end in this error case.

react-backend/routes/postActivity.js
```diff
@@ -1,25 +1,34 @@
...
-      } else {
-        res.send();
-      }
-    });
+      req.on('end', () => {
+        const visitToAdd = JSON.parse(body);
+        console.log(visitToAdd);
+        putVisitsData(visitToAdd.hash, visitToAdd.activity, (error, result) => {
+          if (error) {
+            console.log('I am an error from putVisitsData ', error);
```
ZooeyMiller: again, worth putting a 500 error in here. it's a small thing, and may be tweaked later, but it's better than not send anything to the front end in an error case.

react-backend/routes/postActivity.js
```diff
@@ -1,25 +1,34 @@
...
+          } else {
+            res.send();
```
ZooeyMiller: what does this send? just nothing? (I know you didn't add this code in this PR but it's showed up in the diffs because of the change in indentation,  and it's worth thinking about)

react-backend/app.js
azayneeva: removed this line and the file as we never used it

react-backend/routes/activities.js
azayneeva: added res.status(500).send(err) if it's ok to do error handling like this?

react-backend/routes/getAllUsers.js
azayneeva: here the same: res.status(500).send(err)

react-backend/routes/postActivity.js
azayneeva: done :)

react-backend/routes/postActivity.js
```diff
@@ -23,8 +23,7 @@ router.post('/', (req, res, next) => {
         putVisitsData(visitToAdd.hash, visitToAdd.activity, (error, result) => {
           if (error) {
             console.log('I am an error from putVisitsData ', error);
-          } else {
-            res.send();
+            res.status(500).send(err);
           }
```
ZooeyMiller: We still need to send something back to the front end in the case of a success, even if it's just a message saying "ok" or something similarly trivial.

azayneeva: Sending "success" message now :)

PR #102, opened 2017-12-07
Passwordreset Path

RogeredBacon:
This is the entire password reset path. I've tried to describe exactly what I did in the commit messages.

- Created password reset page where you type your email address and receive an email with your unique token which is generated along with its expiry date each time you request a reset, and stored in two new columns in the cbusiness table.
- You then must navigate to {URL}.com/newPassword and enter this token along with your new password. It will check this is a valid token and it has not expired(1 hour).
- You will then be redirected to the login page.

On my computer atm it seems like the two queries that check if the token is valid + not expired take a long time to complete(on the newPassword page / checkPassword file). This holds up the rest of the validation and makes it seem like it hangs for a bit before redirecting to login. I think this is just an issue with my connection but maybe not?

When you click continue after entering your new password check the console.logs, for me it resolves the first promise instantly then takes a while for the second.

**I haven't properly formatted the emails yet (no /newPassword link, token given twice and messy) so you'll need to manually navigate to /newPassword**


RogeredBacon:
Do we want to have a link to the newpassword page on the login though? I know this would be potentially easier for the user but I feel like we kinda don't want most people to realise the page even exists until they get the email. They'll have to look at the email anyway to be able to get their token. Or does it not really matter because of the security we have?

I've made the token much shorter so it should be easier to copy, we could still do the copy-to-clipboard button but I'm not sure how to do that from with the Postmark website.

Issue #79, opened 2017-11-07
As a CB I can reset my password if I’ve forgotten it.


RogeredBacon:
As a CB I can reset my password if I’ve forgotten it.

- [ ] Creating new pages
- [ ] Implementing email based validation (Reset password email)

PR #109, opened 2017-12-08
Fixed Expired Date.now() bug

RogeredBacon:
Added "Welcome visitor!" message to homeVisitor.js component, put date.now() inside a function


RogeredBacon:
#79 Just fixing the date bug and adding some indication that the visitor sign-up and login is actually aimed at visitors.

PR #110, opened 2017-12-11
Error messages refactoring

azayneeva:
relates to issue #91 

Removed error messages from main.js, signupcb.js, pswdresetcb.js, newPassword.js, logincb.js and added them into new file "errors.js";

replaced "else if" statements for error messages with "switch/case" ones (apart from logincb.js)

### DIFF
--- a/react-backend/client/src/components/authentication/logincb.js
+++ b/react-backend/client/src/components/authentication/logincb.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
 import { Input } from '../visitors/input';
 import { Button } from '../visitors/button';
+import errorMessages from '../errors';
 
 class CBlogin extends Component {
   constructor(props) {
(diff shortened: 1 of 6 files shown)


react-backend/client/src/components/authentication/logincb.js
```diff
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
 import { Input } from '../visitors/input';
 import { Button } from '../visitors/button';
+import errorMessages from '../errors';
```
ZooeyMiller: you could use destructuring here

```js
import { NO_INPUT, EMAIL_ERROR, DETAILS_ERROR } from '../errors';
```

this would mean rather than typing `errorMessages.NO_INPUT` when you use the error message, you could just type `NO_INPUT`. 

I think the way you've done it is fine, just wanted to point this out as an option.

azayneeva: oh right!

PR #113, opened 2017-12-12
added 500 error handling on the front end, relates to issue #17

azayneeva:
added res.status(500).send(err) to backend files where this was missing,
added error handling on the front end for 500
relates to #17

### DIFF
--- a/react-backend/client/src/App.js
+++ b/react-backend/client/src/App.js
@@ -19,6 +19,7 @@ import { HomeVisitor } from './components/visitors/homeVisitor';
 import { HomeAdmin } from './components/admin/homeAdmin';
 
 import { NotFound } from './components/NotFound';
+import { InternalServerError } from './components/InternalServerError';
 
 class App extends Component {
   constructor(props) {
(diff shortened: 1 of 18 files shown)


react-backend/client/src/App.js
```diff
@@ -138,7 +120,7 @@ class App extends Component {
                 )
               }
             />
-
+            <Route exact path="/internalServerError" component={InternalServerError} />
```
ZooeyMiller: you'll need to add this to `frontEndRoutes.js`

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -98,6 +98,11 @@ export class QRCode extends Component {
       headers: this.headers,
       body: JSON.stringify(visitInfo),
     })
+      .then(res => {
+        if (res.status === 500) {
+          throw new Error();
+        }
+      })
       .then(this.props.history.push('/visitor/end'))
```
ZooeyMiller: put this inside of an anonymous function. needs to be `() => this.props.history.push('/visitor/end')`, otherwise you're just directly calling that function.

react-backend/client/src/App.js
azayneeva: Done!

Issue #138, opened 2018-01-09
Add error handling on fetch in adminActivitiesPage.js


azayneeva:
Go back to this pr: https://github.com/FAC-11/DataPower/pull/134

There should probably be some error handling on this fetch. Say if someone lost internet connection, removed some activities and then went to the homepage - currently there wouldn't be any feedback saying something had gone wrong, the activities would just still be there. A better solution would be to grey out the deleted activity in the interim between the removal and the response from the server that says the activity was successfully deleted from the database. Then if the activity wasn't deleted from the server correctly an error could be displayed and the undeleted activity would regain full opacity.

It's an edge case but not all that unlikely.

New issue: IDs should be generated automatically through psql's SERIAL PRIMARY KEY. It's still a good idea to have a temporary random id, but then once you get a response back from the fetch you should replace the id in state with the one generated automatically by the database (make the response the newly created activity so you can grab the actual database id).

So this fetch should have a .then that finds the new activity in state and replaced its randomly generated id with the database generated one. That way there's no possible way to get duplicate activity ids in the database. Like above, there should probably be some indication that the new activity is still waiting to be saved server-side (greyed out or a different colour, etc.)

PR #144, opened 2018-01-10
Update activity ID from database on create

finnhodgkin:
Replaces randomly generated id with database id once an activity has been created.
Also does some quick refactoring of repeated code on the activities page.

#138


RogeredBacon:
oh wait merge conflicts

Issue #120, opened 2018-01-04
 As a CB admin I can sort and filter the data collected through the app


RogeredBacon:
- [x] Tech Spike on filtering - 1 Day
- See data as a list of all the logs made of people coming in, can filter by:      - 1 Day
- [x] Name
- [x] Date of log
- [x] Activity
- [x] Gender
- [x] Age Band
- See data as a list of all visitors that signed up, can filter by:        - 1 Day
- [x] Name
- [x] Date of Sign Up

RogeredBacon:
Done

PR #145, opened 2018-01-10
Sorting for Visits table

azayneeva:
this pr relates to issue #120 
It has functionality for sorting visits table by year of birth, by gender, date of visit, and activity.


react-backend/client/src/components/admin/adminVisitsPage.js
```diff
@@ -67,6 +95,15 @@ export class AdminVisitsPage extends Component {
     return this.state.reauthenticated ? (
       <div>
         <h1>Visitor Data</h1>
+        <select onChange={this.filter}>
+          <option defaultValue value="">
```
finnhodgkin: This default value of an empty string breaks the query if you switch back to it. It should default to the standard sort (date I guess?)

react-backend/client/src/components/admin/adminVisitsPage.js
```diff
@@ -14,20 +14,48 @@ export class AdminVisitsPage extends Component {
     };
   }
 
+  headers = new Headers({
+    Authorization: localStorage.getItem('token'),
```
finnhodgkin: Add a second header of `'Content-Type': 'application/json'` so we can use bodyparser with the request.

react-backend/routes/fetchVisitsFilteredBy.js
```diff
@@ -0,0 +1,23 @@
...
+router.post('/', (req, res, next) => {
+  let body = '';
+  req.on('data', (chunk) => {
+    body += chunk;
+  });
+
+  req.on('end', () => {
+    getVisitsFilteredBy(req.auth.cb_id, body)
+      .then(users => res.send({ success: true, users }))
+      .catch((err) => {
```
finnhodgkin: Swap this to `.catch(next)` to let our error handler in app.js catch it :+1:

react-backend/routes/fetchVisitsFilteredBy.js
```diff
@@ -0,0 +1,23 @@
+const express = require('express');
+
+const router = express.Router();
+
+const getVisitsFilteredBy = require('../database/queries/getVisitsFilteredBy');
+
+router.post('/', (req, res, next) => {
+  let body = '';
+  req.on('data', (chunk) => {
```
finnhodgkin: Once the header is set in the fetch above then this can be removed and replaced with req.body

react-backend/client/src/components/admin/adminVisitsPage.js
azayneeva: it's so weird, but when I do that, the app does nothing - doesn't throw any errors and at the same time doesn't render visits component
when I remover this line - everything starts working again
what can be a problem here?

azayneeva: everything is fine now :)

PR #151, opened 2018-01-11
Filteringvisits

azayneeva:
relates to issue #120 
It has functionality for sorting by gender, age, activity, date of visit and year of birth and filtering by gender, age and activity.

The files are big and probably they need to be modularised somehow, but functionality seems to be working :)


react-backend/client/src/components/admin/adminVisitsPage.js
```diff
@@ -46,9 +82,31 @@ export class AdminVisitsPage extends Component {
...
+  sort = e => {
+    this.setState(
+      {
+        orderBy: e.target.value
+      },
+      this.updateResults
+    );
+  };
+
+  filter = (group, e) => {
```
finnhodgkin: Rather than binding as below, I would prefer this to be partially applied like in my example I descibed before:

```js
filter = group => e => {//..the rest
```

That would mean you don't need to bind below, you just call the function with the group:

```js
 onChange={this.filter('gender')}
```
vs
```js
 onChange={this.filter.bind(this, 'gender')}
```

react-backend/client/src/components/admin/adminVisitsPage.js
```diff
@@ -46,9 +82,31 @@ export class AdminVisitsPage extends Component {
...
+    const filterBy = group + '@' + e.target.value;
+    const isAdding = e.target.checked;
+    let newFilters = [...this.state.filters];
```
finnhodgkin: Here I would prefer:

```js
    const newFilters = isAdding
      ? [...this.state.filters, filterBy]
      : newFilters.filters(filter => filter !== filterBy);
```

react-backend/client/src/components/admin/adminVisitsPage.js
```diff
@@ -92,19 +150,112 @@ export class AdminVisitsPage extends Component {
...
+            </select>
+          </label>
+          <label className="Form__Label">
+            Filter by Gender
+            <br />
+            <label>
+              <input
+                type="checkbox"
+                value="male"
+                onChange={this.filter.bind(this, 'gender')}
```
finnhodgkin: All of these `bind`s can be changed after the above comment about partial application

react-backend/database/queries/getVisitsFilteredBy.js
```diff
@@ -1,27 +1,104 @@
...
-  ORDER BY ${orderBy}`;
-
-const getVisitsFilteredBy = (cbId, orderBy) => {
-  const orderByQuery = {
-    yearofbirth: query('users.yearofbirth'),
-    sex: query('users.sex'),
-    activity: query('activities.name'),
-    date: query('visits.date'),
+const query = (filterBy, orderBy) =>
+  `Select users.id, users.sex, users.yearofbirth, activities.name, visits.date from users inner join visits on users.id=visits.usersid inner join activities on visits.activitiesid = activities.id where activities.cb_id = $1 ${filterBy} ${orderBy}`;
```
finnhodgkin: Try to keep queries readable (capitalised psql commands) and within the 80 character limit: 

```js
const query = (filterBy, orderBy) => `
SELECT users.id, users.sex, users.yearofbirth, activities.name, visits.date
FROM users INNER JOIN visits ON users.id=visits.usersid 
INNER JOIN activities ON visits.activitiesid = activities.id 
WHERE activities.cb_id = $1 ${filterBy} ${orderBy}`
```

react-backend/database/queries/getVisitsFilteredBy.js
```diff
@@ -1,27 +1,104 @@
...
+
+const arrayMatch = (arr, search) => {
+  return arr.some(el => el === search);
+};
+
+const getFiltersQuery = filterBy => {
```
finnhodgkin: I'm going to have at refactoring this function, but for the meantime let's merge and I'll make a separate PR for it.

PR #174, opened 2018-01-29
Morecharts

RogeredBacon:
Added Charts and semi-added bar chart. Merging now so repo is flat before ownership transfer


finnhodgkin:
I think the chart page could be refactored a fair bit but I'm going to merge this so you have a clean slate to work from for the next two days while I'm less present. 

I've also fixed a couple bugs with registering and allowing empty data returns.

Great work guys :+1:

PR #180, opened 2018-02-06
Gmui app bar [Me first]

finnhodgkin:
Adds a material design App bar to all protected routes. Still needs to be added to the unprotected routes, but it should just be a matter of wrapping those routes in a similar component to Private and Admin routes.


eliascodes:
New designs being made, so this is superseded.

PR #191, opened 2018-02-21
Add /api routes

### LINKED ISSUE #173: Add /api to the start of all express routes
Currently our api routes have the potential to conflict with our frontend routes. 
To fix this we should prefix the paths with `/api`.

### LINKED ISSUE #189: Move routes/handlers to their on file
All the routes are hanging out in app.js and it's a bit of a pain to look at. Would be fab to have a router.js file

astroash:
Fixes #189 
Fixes #173 

Potentially a messy fix for adding /api to all routes on the server, happy to manually add /api to all routes if peeps think this is better... 

And my prettier fucked a fair bit up 🙄...

Also I'm terribly out of practise of adding issue numbers to commit messages 🙈

### DIFF
--- a/react-backend/app.js
+++ b/react-backend/app.js
@@ -6,6 +6,7 @@ const cookieParser = require('cookie-parser');
 const bodyParser = require('body-parser');
 const routes = require('./router');
 
+const router = express.Router();
 const app = express();
 
 app.use(favicon(path.join(__dirname, 'public', 'favicon.ico')));
@@ -15,7 +16,8 @@ app.use(bodyParser.urlencoded({ extended: true }));
 app.use(cookieParser());
 app.use(express.static(path.join(__dirname, 'client/build')));
 
-app.get('/api', routes);
+router.use('/api', routes);
+app.use(router);
 app.get('/*', express.static(path.join(__dirname, 'client/build/index.html')));
 
 // catch 404 and forward to error handler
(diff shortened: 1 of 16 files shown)


react-backend/client/src/components/admin/activitiesLib/admin_helpers.js
```diff
@@ -82,4 +78,4 @@ const adminFetch = method => async (that, url, body) => {
 export const adminPost = adminFetch('POST');
 export const adminGet = adminFetch('GET');
 
-export const checkAdmin = that => adminFetch('POST')(that, '/admin/check');
+export const checkAdmin = that => adminFetch('POST')(that, 'api/admin/check');
```
eliascodes: This needs a leading `/` otherwise it will be interpreted as a relative URL. Same for every API route in this PR.

react-backend/app.js
```diff
@@ -45,39 +16,8 @@ app.use(bodyParser.urlencoded({ extended: true }));
...
-app.use('/visitors/all', mwAdminIsAuthenticated, visitorsAll);
-app.use('/visitors/filtered', mwAdminIsAuthenticated, visitorsFiltered);
-app.use('/admin/check', mwAdminIsAuthenticated, adminCheck);
-app.use('/activities/all', mwAdminIsAuthenticated, activitiesAll);
-app.use('/activity/add', mwAdminIsAuthenticated, activitiesAdd);
-app.use('/activity/update', mwAdminIsAuthenticated, activitiesUpdate);
-app.use('/activity/delete', mwAdminIsAuthenticated, activitiesDelete);
-
+router.use('/api', routes);
+app.use(router);
```
eliascodes: I think you're creating an unnecessary router here. Can't you just do `app.use('/api', routes)`?

Also probably better to call it `apiRoutes` or something like that.

Issue #218, opened 2018-02-25
Write tests for everything ever


RogeredBacon:
We need tests for everything; this is the issue

astroash:
Relates #89

eliascodes:
Not sure how this is more specific than #89 (or #127), but at this stage I'll let it go 😅

astroash:
@RogeredBacon can this issue be closed?

RogeredBacon:
Has everything ever been tested?

astroash:
Fair point.

eliascodes:
This can never be closed, so I'm closing it.

PR #219, opened 2018-02-25
[WIP] Added tests for /api/user/name-from-scan route.

RogeredBacon:
I've been going through the non-admin authenticated routes in the router.js.

Because the route is authenticated I'm getting jwt related errors when I try to run the tests. I need to generate a jwt, I think, to be able to access the routes.

I'm not confident with testing and would appreciate it if you guys could have a look and tell me how I would go about fixing the issue. Once this is sorted I'll apply the fix to my other tests and hopefully that should be all of the non-admin authenticated routes tested.

#218


tests/routes_tests/visitor_name.test.js
```diff
@@ -0,0 +1,56 @@
...
+    .expect(res => res.json().fullname)
+    .end((err, res) => {
+      t.ok(res === 'james bond', 'Hash is viable and found in database');
+      t.end();
+    });
+});
+
+test('POST /api/user/name-from-scan | non-viable hash ', (t) => {
+  const failPayload = {
+    hashToCheck: 'Im faily sure, though not positive, that this is not a hash',
```
eliascodes: 😆

RogeredBacon:
I've added the token as you said in (1) which has solved the jwt warnings I was getting. However I'm getting 500 errors now, even when I'm not calling the database.

I think this might have something to do with the mw_is_authenticated.js file and its query, though I'm not sure. Would you please mind having a look and telling me what you think.

RogeredBacon:
>  POST /api/user/name-from-scan 500 37.367 ms - 1420

>       operator: ok
>       expected: true
>       actual:   false
>       at: Test.request.post.set.send.expect.expect.expect.end (/home/rogeredbacon/Documents/DataPower/tests/routes_tests/visitor_name.test.js:21:9)
>       stack: |-

^ Is the only logging I can see. I'm not sure what's causing it, but when I change the credentials in the jwt, to be wrong, I can see the error message in the reject you mentioned, 'no results found'. It seems like otherwise the query is working fine?

tests/routes_tests/visitor_name.test.js
```diff
@@ -0,0 +1,66 @@
...
+  };
+  request(app)
+    .post('/api/user/name-from-scan')
+    .set('authorization', token)
+    .send(successPayload)
+    .expect(200)
+    .expect('Content-Type', /json/)
+    .expect(res => res.json().fullname)
+    .end((err, res) => {
+      t.ok(res === 'james bond', 'Hash is viable and found in database');
```
eliascodes: Instead of using `t.ok` to check if two things are equal, you should use the provided `t.equal`. It's more readable, and gives you more useful output in case of failure compared to the `expected: true, actual: false` you're getting now.

tests/routes_tests/visitor_name.test.js
```diff
@@ -0,0 +1,66 @@
...
+  };
+  request(app)
+    .post('/api/user/name-from-scan')
+    .set('authorization', token)
+    .send(failPayload)
+    .expect(400)
+    .expect('Content-Type', /json/)
+    .expect(res => res.json().fullname)
+    .end((err, res) => {
+      t.notOk(res === 'Bad hash', 'Validation returned output of "Bad Hash"');
```
eliascodes: Similarly, `t.notEqual` instead of `t.notOk`

tests/routes_tests/visitor_name.test.js
eliascodes: When debugging, your job is to trace the problem back from where it appears to where is has originated from. Your instinct should always be to think about how to try and get more information on what is being returned or calculated at every step. 

In your case, your assertion is expecting `true`, but receiving a `false`. This is because the comparison fails. Your first port of call should be to look at what you're actually getting back in `res` and see what that tells you about where in the backend code the response must have come from.

RogeredBacon: ah thanks, I'll do that. Though now I think I'm getting too much information haha

astroash: Just to join in, I feel like line 19 should be completed in the `.end` to be more readable  . It feels weird using `.expect` like middleware.

RogeredBacon: Yeah turns out it wasn't actually doing anything anyway so I've taken it out

RogeredBacon:
Well, I've got them working locally but now travis is saying that that there's a relation error with the database. God dammit

eliascodes:
@RogeredBacon your test database wont be build automatically, you need to actually make sure that happens yourself. Checkout the tests that Finn added (specifically the `database_rebuild.js` file).

tests/getVisitsFilteredBy.test.js
```diff
@@ -149,34 +149,34 @@ test('Sorts by field', async (t) => {
...
+    const filterByMale = await visitorsFiltered(1, {
       filterBy: ['gender@male'],
     });
 
-    const sortByBadValue = await visitorsFiltered(2, {
+    const sortByBadValue = await visitorsFiltered(1, {
       orderBy: 'badvalue',
     });
 
     t.deepEqual(sortByYear, sortedByYear, 'Sorts by year');
```
RogeredBacon: This test here is not working. I'm getting back what I want however the 'sortedByYear' function seems to just be returning the results in a random order rather than sorted and I'm not sure why. So the 'actual is correct however the 'expected' is not.

RogeredBacon:
It seems like the ENV variables are causing grief, Travis is failing because of the postmark secret. Is this what you mentioned earlier @eliascodes ?

Issue #221, opened 2018-02-26
Test database queries


astroash:
Test remaining database queries

astroash:
FYI, I’m still working on this. But I stopped being annoying and refactoring all the routes they touch 😊.

astroash:
One query left untested 🙈

astroash:
🏰 and forever it shall stay...

astroash:
Closing as we are migrating to a new db

PR #223, opened 2018-02-26
Moar db tests

astroash:
- Refactor and test of visitor check
- Rename functions and files
- Remove pointless promise
- Add status codes
- Test db userCheck Exists

I did some file renaming but there's still 2 _conventions_ going on, routes name this `visitor_check` and db calls it `user_check_exists`. I didn't want to change either one as they're grouped with other similar files in their folders.

 #221


eliascodes:
And then similarly `routes_tests` can just be `routes` (didn't spot that in the first PR)

astroash:
Oops sorry, I'm gunna continue on this branch for today.

react-backend/client/src/components/visitors/main.js
```diff
@@ -43,7 +43,7 @@ export default class Main extends Component {
     'Content-Type': 'application/json',
   });
 
-  checkUserExists = (e) => {
+  userCheckExists = (e) => {
```
eliascodes: Why is this a better name than before? Before the method name began with a verb, and had an instructional voice. The new one sounds like the user is doing something, which is not the case.

astroash: Because it matches the db query functions/files. Lots of similar ones are prefixed with _user_ so didn't want to change the word order.

Let me know if you think it's actually better the previous way and I can change it back. Just trying to create some order 😕

eliascodes: Don't think we need to match method names in the front end with file name in the backend. That's too much coupling IMO. The contract is the API.

astroash: @eliascodes not sure I'm following... Won't this make me loose the work that I've done since? Or will I need to stash this first then `git reset [branch]`?

eliascodes: If you have done work since, yes you will lose those commits (although you will keep the changes in your working copy as long as you don't use `--hard`). tbh not a big deal either way, just tidier to not have commits that are not meaningful. You can always solve it after the fact with an interactive rebase :wink:

PR #224, opened 2018-02-27
App configuration re-organisation

### LINKED ISSUE #190: App configuration
Currently all configuration of the application is done through the `config.env` file, and there are various checks and `if` statements with references to `process.env` throughout the codebase.

This is because the `config.env` file is currently being used for two things:
1. to hold secrets
2. to hold application configuration

But these things should be dealt with separately. The application can ingest secrets and top-level environment flags via environment variables (for example passwords or encryption secrets), and all other configuration options can be stored in one or more JSON files.

One configuration component will read both the environment variables and the JSON config files, normalise them all into one structure, and _this_ is the configuration that is to be referred to throughout the application.

This will mean all configuration validation logic can be kept in one place, and you don't need to come up with semantically ambiguous environment variable names to differentiate values that will be used in different environments (e.g. apparently we have `DATABASE_URL` and `DATABASE_TEST`).

Suggest using [`nconf`](https://www.npmjs.com/package/nconf) instead of rolling our own.

### LINKED ISSUE #220: Utility library preference
I notice there's not yet a utility library in use on the project. I find it useful to have one usually as it prevents me from having to write too much of my own code that does stuff like object transformations or merging.

I'm aware of and have used [ramda](http://ramdajs.com/), [lodash](http://lodash.com/) and [underscore](http://underscorejs.org/).

Thoughts on using one of these? Or do you have another preference? There's no compulsion to use it, just available to do tedious things if you want it.

eliascodes:
Fixes #190 
Fixes #220

So here it is!

### Main changes
* We now have a single configuration object that should be referred to by all code
* `app.js` now exports a function that creates an express app given a configuration object
* This means tests relying on the `app` now need to create it themselves with the appropriate configuration. I've done this for the tests that are already on master.
* This is actually only a partial update because currently `dbConnection.js` and all files relying on it are unchanged. I decided to leave this because it would inflate the diff unnecessarily. I will finish off this work in a subsequent PR. After that's done we should be able to remove the `process.exit` nonsense from the tests 🎉 

### Minor changes
* `ramda` and `joi` added as dependencies
* Removed one of the nested `eslintrc` files

### DIFF
--- a/react-backend/app.js
+++ b/react-backend/app.js
@@ -4,43 +4,37 @@ const favicon = require('serve-favicon');
 const logger = require('morgan');
 const cookieParser = require('cookie-parser');
 const bodyParser = require('body-parser');
-const apiRoutes = require('./router');
+const postmark = require('postmark');
+const createPgPool = require('./database/dbConnectionLazy');
+const { notFound, errorHandler } = require('./shared/middleware');
 
-const app = express();
 
-app.use(favicon(path.join(__dirname, 'public', 'favicon.ico')));
-app.use(logger('dev'));
-app.use(bodyParser.json());
-app.use(bodyParser.urlencoded({ extended: true }));
-app.use(cookieParser());
-app.use(express.static(path.join(__dirname, 'client/build')));
+module.exports = (cfg) => {
+  // Non-global import required because of client-connections
+  // that are created in module scope of routes
+  const apiRoutes = require('./router'); // eslint-disable-line global-require
 
-app.use('/api', apiRoutes);
-app.get('/*', express.static(path.join(__dirname, 'client/build/index.html')));
+  const postmarkClient = new postmark.Client(cfg.email.postmark_key);
+  const psqlClient = createPgPool(cfg);
 
-// catch 404 and forward to error handler
-app.use((req, res, next) => {
-  const err = new Error('Not Found');
-  err.status = 404;
-  next(err);
-});
+  const app = express();
 
-// error handler
-app.use((err, req, res) => {
-  if (err === 'notauthorized') {
-    return res.status(401).send({ error: 'Not logged in' });
-  }
-  // set locals, only providing error in development
-  const message = err.message;
-  const error = req.app.get('env') === 'development' ? err : {};
+  app.set('cfg', cfg);
+  app.set('client:postmark', postmarkClient);
+  app.set('client:psql', psqlClient);
 
-  // render the error page
-  res.status(err.status || 500);
-  res.send({ error, message });
-});
+  app.use(favicon(path.join(__dirname, 'public', 'favicon.ico')));
+  app.use(logger(cfg.env));
+  app.use(bodyParser.json());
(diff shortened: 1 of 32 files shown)


react-backend/app.js
```diff
@@ -4,43 +4,37 @@ const favicon = require('serve-favicon');
...
+  const apiRoutes = require('./router'); // eslint-disable-line global-require
+
+  const postmarkClient = new postmark.Client(cfg.email.postmark_key);
+  const psqlClient = createPgPool(cfg);
+
+  const app = express();
+
+  app.set('cfg', cfg);
+  app.set('client:postmark', postmarkClient);
+  app.set('client:psql', psqlClient);
```
astroash: This feels a bit gross, why are we doing this? 🙈

astroash:
It's a bit OOP, innit mate?

react-backend/app.js
eliascodes: Why is it gross?

Storing the clients on the app object means they can be accessed from the handlers without needing to require anything in. That way they can be passed as arguments to functions that need them.

eliascodes:
> It's a bit OOP, innit mate?

Which aspect of it is OOP? I can assure you it's significantly more functional now than it was before.

react-backend/app.js
astroash: It feels like a weird separation of concerns thing having access on each route. Would we be accessing the dB from the request from now on?

eliascodes: Not sure what you mean by "weird separation of concerns thing", nor exactly what the objection to this is, but we can talk about it during our chat tomorrow.

Yes, handlers would access database clients via `req.app` and then pass it on as an argument to whatever functions needed it.

eliascodes: Think of it this way: before, handlers and functions below them in the hierarchy were relying on object references hauled in from outside using `require`. Now, they only rely on their arguments.

astroash: I'm not objecting in the sense that I don't think it should be merged, just that it seems strange compared to the _limited setups worked with_. All the other config middleware you've added seems to add variables that we can access them where needed, whereas this is adding a db pool just incase we need it. 

I guess I don't really see where the separation is from what should be added to the request through middleware. Technically we could add all our functions as middleware too and only ever require things once, but that would be mental...

Also  😭,updating Travs envs has broken my tests, so let's get this merged first 😊.

eliascodes: > Also 😭,updating Travs envs has broken my tests

Really sorry, I wasn't thinking when I did that. You can just add your env vars back to the config, and it should work.

> whereas this is adding a db pool just incase we need it

Ah, I think there's been  a misunderstanding. I'm not adding the pool just in case we need it. We _will_ need it. Just not in this PR. See my PR description:

> This is actually only a partial update because currently dbConnection.js and all files relying on it are unchanged. I decided to leave this because it would inflate the diff unnecessarily. I will finish off this work in a subsequent PR.

In the next PR, I'm going to remove all references to `dbConnection.js`.

Issue #234, opened 2018-02-28
Name consistency for 3 types of users


astroash:
We have 3 different types of users for this app and they are often called different things. It would be great to have a consistent name for each. It will make naming conventions in the app easier and simpler for us to talk about. As far as I am aware the different names are:

users/visitors
cb/cbusiness/admin
superadmin/twine

I would be up for calling them:
users
cbusiness
admin

But not sure if that accurately covers the use case for each.

eliascodes:
🙏 thanks for opening this

My vote is for:
* visitors
* cbusinesses
* admins

> But not sure if that accurately covers the use case for each.

👆 this is important, maybe @RogeredBacon can fill us in 😃

astroash:
I've decided to go for admin/jwt_secret for the env variables as the the `jwt_secret` is used on both types of users depending on their flow.

eliascodes:
OK but can it be `jwt_secret` and `admin_jwt_secret`. Just calling it `admin_secret` doesn't tell anyone what it's used for.

astroash:
Yeah, sorry that's what I've gone for. Poor use of a forward slash 😞

eliascodes:
What about Visitor/CbAdmin/TwineAdmin? 🔧🙈

astroash:
☝️ Ohhh, I _think_ this works...

@RogeredBacon what do you think?

RogeredBacon:
> What about Visitor/CbAdmin/TwineAdmin? 

I think this one is the clearest, any ambiguity would only come from the fact that you need to be logged in as a CbAdmin to use the app at all really (except registering/reset password etc.), then you'll need to refresh your credentials to be able to access the CbAdmin dashboard. The difference between:

`jwt = { cbEmail: foo }` and `jwt = { cbEmail: foo, admin: true }`

astroash:
Sweet, this should be added somewhere in the docs 😊

PR #236, opened 2018-02-28
🛠 Fix references to jwt secret to include admin version

### LINKED ISSUE #235: Incorrect env variables used for jwt
Depending on the flow of a cbusiness user, they are signed in with a different jwt secret. This was accidentally removed in #224 

If an cbusiness user logs in on the front page they are signed with a normal `jwt_secret`.
If they sign in to their own admin section they are signed with an `admin_jwt_secret`. Login to this area of the app comes with a 5 minute refresh token.

astroash:
- add eslint magic
- update docs to reflect two types of jwt
- include admin_jwt_secret for when cb's login to their admin section

I think I got them all, but let me know if you think anything is missing 😊

Fixes #235 
 #234 #224


docs/config_guide.md
```diff
@@ -1,7 +1,8 @@
 # Configuration Guide
+
 All configuration is stored in the `config` directory located in the project root.
 
-Secrets and configuration variables that are likely to change between deployment environments are stored in a file called `config.env`. This file should *NOT* be checked into source control.
+Secrets and configuration variables that are likely to change between deployment environments are stored in a file called `config.env`. This file should _NOT_ be checked into source control.
```
eliascodes: Looks like someone hasn't stopped prettier from linting markdown files 🤔 lol

eliascodes:
@astroash are we sticking with admin == cbusiness? 😕

docs/config_guide.md
astroash: 🙈 I turned it off half way through this PR

astroash:
> @astroash are we sticking with admin == cbusiness? 😕

@eliascodes there's actually two use cases for the standard `jwt_secret`, it's not just that one is used for cbusinesses and one is used for visitors. I've editted #235 with more details. Let me know if it doesn't make sense.

eliascodes:
@astroash I was talking about the naming convention. You've called the key `admin_jwt_secret`, not `cbusiness_jwt_secret`.

astroash:
@eliascodes yaha, I know that's what you were asking. `jwt_secret` is used on both visitors and cbusiness. cbusinesses have their own admin section to the app, which is when they use a different jwt (`admin_jwt_secret`).

eliascodes:
Lol, but `admin_jwt_secret` will only be used by cbusinesses. If we then have real admin users, what are we going to call their JWT secret?

astroash:
I reckon it's misleading if we call it `cbusiness_jwt_secret`, and have a section on the front page where cbusinesses login with `jwt_secret` not the `cbusiness_jwt_secret`. Maybe they could be called `standard_jwt_secret`, `cbusiness_admin_jwt_secret` & `twine_admin_secret` (for when we add the third flow)?

eliascodes:
> `standard_jwt_secret`, `cbusiness_admin_jwt_secret` & `twine_admin_secret`

That's closer. See [my comment](https://github.com/TwinePlatform/DataPower/issues/234#issuecomment-369298163)

PR #247, opened 2018-03-05
Password reset email now sets token value automatically + CB name validation changed to alphanumeric

### LINKED ISSUE #117: Fix validation for CB name when signing up
Currently its not allowed to input any non-alphabetical characters.

### LINKED ISSUE #128: Need to fix the password reset email to make the link automatically input the token

RogeredBacon:
Changed password template and newPassword page to automatically set the token string to state, meaning the user can just click the link and pick a new password.

Changed CB name validation to alphanumeric rather than alpha

Fixes #128
Fixes #117


react-backend/client/src/components/authentication/newPassword.js
```diff
@@ -8,11 +8,12 @@ import errorMessages from '../errors';
 export default class NewPassword extends Component {
   constructor(props) {
     super(props);
+    const token = this.props.match.params.token;
```
eliascodes: IMO it's safer to move this to `componentWillMount` instead of the `constructor`. As far as I understand, the `constructor` will be called only _once_ for the whole lifetime of the app, whereas `componentWillMount` will be called each time the component is mounted onto the DOM.

react-backend/routes/cbauthentication/cb_register_check.js
```diff
@@ -19,7 +19,7 @@ router.post('/', (req, res, next) => {
   const orgName = formName.split(' ').join('');
 
   const notEmail = !validator.isEmail(formEmail);
-  const notLatinName = !validator.isAlpha(orgName, ['en-GB']);
+  const notLatinName = !validator.isAlphanumeric(orgName, ['en-GB']);
```
eliascodes: While we're here, we may as well add a test for this route :wink:

PR #248, opened 2018-03-05
[WIP] Started adding validation for CBSettings and UserDetails pages

### LINKED ISSUE #161: Need to have input validation for usersDetailsPage and cbSettingsPage
At the moment it just checks for no-input, we just need the same sort of thing we do in the cb signup page

RogeredBacon:
Started adding validation on the routes associated with adminCBSettingsPage and adminUserDetailsPage along with error messages and stuff similar to the CBSignup page.

Fixes #161 

*This will probably cause some tests to fail/become irrelevant so I'll fix/add appropriate tests when it all works

### DIFF
--- a/react-backend/client/src/components/admin/adminCBSettingsPage.js
+++ b/react-backend/client/src/components/admin/adminCBSettingsPage.js
@@ -4,6 +4,7 @@ import Dropzone from 'react-dropzone';
 import { Link } from 'react-router-dom';
 import Logoutbutton from '../visitors/logoutbutton';
 import { adminPost } from './activitiesLib/admin_helpers';
+import errorMessages from '../errors';
 
 export default class AdminCBSettingsPage extends Component {
   constructor(props) {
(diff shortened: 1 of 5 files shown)


react-backend/client/src/components/admin/adminCBSettingsPage.js
```diff
@@ -135,7 +138,36 @@ export default class AdminCBSettingsPage extends Component {
...
+            this.setError([
+              errorMessages.NAME_ERROR,
+              errorMessages.EMAIL_ERROR,
+            ]);
+            break;
+          case 'noinput':
+            this.setError([errorMessages.NO_INPUT_ERROR]);
+            break;
+          default:
+            return data.details;
```
eliascodes: this is redundant if you've got `return` on line 169, you can just put `break` here

eliascodes:
Oops, just seen:

> ...so I'll fix/add appropriate tests...

👍

PR #249, opened 2018-03-06
📝 add note on postmark API key for dev/test environments

### LINKED ISSUE #240: Issue with rejected promises from postmark
After implementing #238, it good old trav has flagged an issue as tests are failing. I am going to do some digging to see what's going wrong...

Relates #239

eliascodes:
Fixes #240

@astroash @RogeredBacon Make sure you've set `POSTMARK_KEY_DEV` and `POSTMARK_KEY_TEST` to equal `POSTMARK_API_TEST`.


astroash:
@eliascodes I reckon we want to be able to send emails in dev environment, just not in test. When logging into the app to manually test, I've only ever used my actual email (or James) rather than the ones in `data.sql`. Having our app actually emailing allows you to get a new copy of the qr code and reset emails ect.

RogeredBacon:
Good point, I agree

docs/config_guide.md
```diff
@@ -24,6 +24,9 @@ DATABASE_URL_TEST=...
 
 # Postmark API key for various environments
 # Postmark is the service used to send e-mails
+# NOTE: Use "POSTMARK_API_TEST" as the API key in "test" and "dev" environments, this will prevent sending
```
astroash: Can you remove "dev" environments? Also is it worth linking to postmarks docs rather than our issue?

eliascodes: The bit about the test API key isn't totally obvious in the postmark docs, our issue has a screenshot of the right part of the docs.

Issue #255, opened 2018-03-07
As a visitor, I can give feedback through the main page by...


astroash:
selecting 3 smileys titled- “How did we do today?”

- [x] Create a new table in database to collect feedback #256 - *1.5 days*
- [x] Create front end view with buttons that update the database - *1 day*
- [x] Create front end view for cb-admin to see their feedback #280 - *2 days*
- [ ] ~~Create front end view for twine-admins to see all cbs feedback #304 - *2 days*~~

Issue #244, opened 2018-03-05
Inconsistent way of handling incorrect dB queries


astroash:
Some db queries return an empty array if they can’t find what they’re looking for, whereas other return a rejected promise. Moving forward to would be good to pick one consistent way of doing this 😊

RogeredBacon:
Yeah we need to overhaul all of the routes and queries to use a uniform way of returning results/errors. Currently there's like 4/5 different types of response going on.

eliascodes:
IMO:

Since promises represent future values, they should reject only if it is impossible to return a value. In JS, we live a world where we have a "value" (`null`) that represents "no-value". Think about how `pg` uses promises: if no records are found, the promise doesn't reject, it resolves with an "empty" value (in most cases an empty array in `result.rows`).

There are two cases when thinking about return values:
1. Queries that return 0-1 record(s)
   * These promises represent a single record, so should resolve with `null` if no records found.
2. Queries that return 0-Infinity records
   * These promises represent collections of records, so should always resolve with an array. If no matching records are found, they should simply resolve with an empty array.

With this convention, we can be sure that any rejected promises represent serious errors outside of the normal flow of application logic. Otherwise we'd need logic inside `catch` to determine if a promise was rejected because something went wrong, or just because no record was found.

astroash:
I like the idea of not unnecessarily rejecting promises, but you've lost me a little on points 1 & 2. 

In your description, what's the difference between **point 1** returning `null` if _"no records found"_ and **point 2** returning `[]` if _"no matching records found"_.

eliascodes:
> In your description, what's the difference between point 1 returning null if "no records found" and point 2 returning [] if "no matching records found".

Think about the difference between:
1. "Find the user with an id of 2"
2. "Find all users with the name 'John'"

astroash:
Lol, that's a very complex description for wether or not our result needs to return an array.

astroash:
@eliascodes following on from above. Do you think we should return an `Error` or `null` for a query that is trying to `INSERT` data into the db?

eliascodes:
@astroash that would cause an error in SQL so IMO the promise should be rejected. It's indicative of a programmer error, which we should not try to recover from.

astroash:
@eliascodes do you think this needs to be outlined somewhere in the docs? I _think_ we only mention expected data for apis atm.

eliascodes:
[...]

PR #262, opened 2018-03-08
🗃 Feedback Flow

### LINKED ISSUE #256: Create a new table in database to collect feedback
With tests and all ❤️

astroash:
👀Please REVIEW me👀

**Disclaimer:**  I've been staring at this feature for too long so 🤷🏽‍♀️

**Breakdown**
- Attempt some architecture nonsense
- Update db files to include creating/destruction of new feedback table
- Create new route for /api/cb/feedback
- Create generalise `SELECT` query to take date ranges
- Create `INSERT` query for feedback magic
- Add feedback buttons to frontend
- Add thank you component with redirect magic
- 👷‍♂️ ALL OF THE TESTS 👷
- 📝 And even some docs

#255 
Fixes #256


astroash:
@eliascodes is this what you were imagining with your [architecture plan](https://github.com/TwinePlatform/DataPower/blob/master/docs/architecture.md)?

db/schema.sql
```diff
@@ -44,10 +44,18 @@ CREATE TABLE activities
...
   activitiesId INTEGER REFERENCES activities(id),
   cb_id INTEGER REFERENCES cbusiness(id),
   date TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP
 );
 
+CREATE TABLE feedback
+(
+  id SERIAL PRIMARY KEY,
+  cb_id INTEGER REFERENCES cbusiness(id),
+  user_feedback INTEGER NOT NULL,
```
eliascodes: 🙏 please can we destroy as many references to "user" as possible, IMO it's nothing but confusing. IMO `feedback_score` is better because we are collecting quantitative feedback, not qualitative.

react-backend/shared/models/cb-admin.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  getCbFeedback: async (dbConnection, cbId) => {
```
eliascodes: If you imagine that this object is describing the `CbAdmin` model, then putting `CB` in the method name seems redundant; there are no other kinds of feedback we can collect.

```js
const feedback = await CbAdmin.getFeedback(client, id);

// compared to ...

const feedback = await CbAdmin.getCbFeedback(client, id);
```

Separate note (with no action right now): I usually like to make these methods as general as possible to capture future use cases more easily. For example, at the moment, this query can only give us all the feedback for a CbAdmin. But what if we want to filter by date in future? Wouldn't it be nice to be able to do something like:

```js
const then = new Date(Date.now() - 1e10)
const now = new Date(Date.now())
const feedback = await CbAdmin.getFeedback(client, { id, since: then, until: now });
```

tests/shared/models/cb-admin.test.js
```diff
@@ -0,0 +1,64 @@
+const test = require('tape');
+const pg = require('pg');
+const { getConfig } = require('../../../config');
+const { refresh: refreshDB } = require('../../../db/scripts');
+const cbAdmin = require('../../../react-backend/shared/models/cb-admin');
+
+const config = getConfig(process.env.NODE_ENV);
+test.only('Db Query | cbAdmin.getCbFeedback ', async tape => {
```
eliascodes: watch out for this `only`

react-backend/shared/models/cb-admin.js
astroash: Oooooo

db/schema.sql
astroash: Yeah, makes sense 😊

tests/shared/models/cb-admin.test.js
astroash: Come on, its a [WIP]...

astroash:
> Yeah this is pretty much along the lines of what I was thinking.

👌

react-backend/shared/models/cb-admin.js
```diff
@@ -0,0 +1,19 @@
+const cbAdmin = {
+  then: new Date('2000-01-01T00:00:00.000Z'),
+  now: new Date(Date.now()),
+
+  getFeedback: async (
+    dbConnection,
+    { cbId, since = cbAdmin.then, until = cbAdmin.now }
```
astroash: @eliascodes, the internet tells me I can't do this but it works so 🤷🏽‍♀️...

react-backend/client/src/components/home.js
```diff
@@ -1,11 +1,29 @@
...
+      feedbackScore,
+    }),
+  });
+};
 const Home = () => (
   <div>
     <h1>
       Who are you?<br />
     </h1>
+    <button onClick={() => postFeedback(1, -1)}>-1</button>
```
eliascodes: where are these id's coming from?

react-backend/shared/models/cb-admin.js
```diff
@@ -0,0 +1,32 @@
+const then = new Date('2000-01-01T00:00:00.000Z');
+const now = new Date(Date.now());
+
+module.exports = {
+  getFeedback: async (dbConnection, { cbId, since = then, until = now }) => {
+    const getFeedbackQuery = `SELECT feedback_score, feedback_date FROM feedback WHERE cb_id = $1 AND feedback_date BETWEEN $2 AND $3`;
```
eliascodes: Is there an endpoint where this query is exposed?

react-backend/shared/models/cb-admin.js
```diff
@@ -0,0 +1,32 @@
...
+    const query = await dbConnection.query(getFeedbackQuery, [
+      cbId,
+      since,
+      until,
+    ]);
+    return query.rows;
+  },
+
+  postFeedback: async (dbConnection, { cbEmail, feedbackScore }) => {
```
eliascodes: Get rid of `post` here. The DB layer is not related to the API layer. There's no reason to introduce a coupling in the naming of methods and the endpoints on which they're used. My use of `get` in `getFeedback` was not a reference to the HTTP `GET` method. 

Instead, name the method for what it does in the context of the model. It adds a new feedback record. What might be a better name?

react-backend/shared/models/cb-admin.js
```diff
@@ -0,0 +1,32 @@
...
+        WHERE cbusiness.email = $1 LIMIT 1)
+    , $2) 
+    RETURNING *`;
+
+    const query = await dbConnection.query(postFeedbackQuery, [
+      cbEmail,
+      feedbackScore,
+    ]);
+
+    return query.rows;
```
eliascodes: If this query is returning only one record, why does it return an array? Remember #244

react-backend/client/src/components/home.js
```diff
@@ -1,11 +1,29 @@
...
+  });
+
+  return fetch('/api/cb/feedback', {
+    method: 'POST',
+    headers,
+    body: JSON.stringify({
+      cbId,
+      feedbackScore,
+    }),
+  });
```
eliascodes: Then what? How does the visitor know that their feedback was successfully or unsuccessfully recorded?

tests/shared/controllers/feedback.test.js
```diff
@@ -0,0 +1,55 @@
...
+      const token = jwt.sign({ email: 'fakey@mcfake.com' }, secret);
+      const badPayload = {
+        feedbackScore: -1,
+      };
+
+      request(app)
+        .post('/api/cb/feedback')
+        .set('authorization', token)
+        .send(badPayload)
+        .expect(500)
```
eliascodes: Why does a bad payload result in a `5xx`? `4xx` errors denote problems with the request.

tests/shared/models/cb-admin.test.js
```diff
@@ -0,0 +1,144 @@
...
+          feedback_date: new Date('2017-12-09T17:12:00.000Z'),
+        },
+      ];
+
+      t.deepEquals(
+        actualAll,
+        expectedAll,
+        'cbAdmin.getFeedback returns all cbs feedback without dates specified'
+      );
+      const sinceDate = new Date('2017-11-09 17:23:00+00');
```
eliascodes: Split these with/without dates into separate test cases.

docs/api_migration_plan.md
```diff
@@ -82,3 +82,4 @@ All non-successful responses should return appropriate status codes in the range
 | `POST   /cbs/login` | | `{ query: { email, password } }` | `{ result: { cb_admin_token } }` | `POST /admin/login` |
 | `POST   /cbs/:id/emails` | | `{ query: { password_reset: Bool } }` | `{ result: null }` | `POST /cb/pwd/reset` |
 | `POST   /visitors/statistics` | `Authorization: cb_admin_token` | `{ query: { group_by: { gender, activity, age: [18, 35, 51, 70], days: ['monday'] } }` | `{ result: { gender: { male, female, ... }, activity: { yoga: 12, ...}, age: { 0-17: 2, ... }, days: { monday: 2, ... } } }` | `POST /users/chart-all` <br/> `POST /users/filtered` |
+| `POST   /cb/feedback` | `Authorization: cb_admin_token` | ` { email }` | `[ { cbEmail, feedbackScore, feedback_date } ]`| ✨ shiney shiney new ✨ |
```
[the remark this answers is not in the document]
astroash: It depends how shiny it is....

react-backend/shared/controllers/feedback.js
```diff
@@ -0,0 +1,21 @@
+const jwt = require('jsonwebtoken');
+const cbAdmin = require('../models/cb-admin');
+
+const route = ('/',
```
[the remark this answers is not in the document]
astroash: A function...

Your `x` is just a bunch of numbers wrapped in parens...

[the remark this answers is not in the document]
eliascodes: The value of `x` in my example is `3`. I was pointing out that assigning `(x, y, z)` to a variable is redundant, because `x` and `y` are ignored. You are assigning `('/', () => {...})` to `route`.

react-backend/shared/models/cb-admin.js
```diff
@@ -0,0 +1,32 @@
...
+  },
+
+  postFeedback: async (dbConnection, { cbEmail, feedbackScore }) => {
+    const postFeedbackQuery = `
+    INSERT INTO feedback 
+    (cb_id, feedback_score) 
+    VALUES (
+      (SELECT cbusiness.id FROM cbusiness 
+        WHERE cbusiness.email = $1 LIMIT 1)
```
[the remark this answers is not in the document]
astroash: Na, I think this is left overs of a bizarre path I stumbled down trying to inner join on feedback n only match with emails of previously entries 😅

react-backend/shared/models/cb-admin.js
```diff
@@ -0,0 +1,32 @@
+const then = new Date('2000-01-01T00:00:00.000Z');
+const now = new Date(Date.now());
```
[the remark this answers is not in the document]
astroash: Oops, yeah, I guess setting it to a function returning `Date.now()` instead?

react-backend/client/src/components/home.js
astroash: Yup, that explains the overwhelming feeling that I forgot something 😊

react-backend/shared/models/cb-admin.js
astroash: Not yet.

react-backend/client/src/components/home.js
RogeredBacon: Are we gonna restrict how often feedback can be given? Some sort of waiting period, to stop people spamming. Maybe an issue for another time.

react-backend/shared/models/cb-admin.js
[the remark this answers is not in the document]
RogeredBacon: Good for safety but our cb_signup validation should do this anyway

react-backend/client/src/components/home.js
astroash: ☝️ Yeah, that was kinda my thinking. _Vaguely_ css related...

eliascodes: Great shout @RogeredBacon. A frontend-only approach to abusive/malicious use cases is not a solution. Needs additional backend logic

react-backend/shared/controllers/feedback.js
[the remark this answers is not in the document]
astroash: Lol, oh yeah, I didn't realise I was doing that.

react-backend/shared/models/cb-admin.js
```diff
@@ -0,0 +1,32 @@
...
+
+module.exports = {
+  getFeedback: async (dbConnection, { cbId, since = then, until = now }) => {
```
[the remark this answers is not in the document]
astroash: ## ☝️Option 3 ☝️ 
This was the **woes** I was talking about yesterday. I do not have access to the cb id in the front end where this query is called from, just the cb's email. I wanted to store the `cb_id` in the `feedback` table cos it's a foreign key, so I'm grabbing the data with the id, but have to insert & match on email address.

react-backend/shared/models/cb-admin.js
astroash: Lol, I cannot help but read this in a primary school teacher voice...

react-backend/client/src/components/home.js
astroash: Oops, okay, I should have looked at the designs better, this is actually all on the wrong page and should have a redirect on the client... brb..

tests/shared/models/cb-admin.test.js
astroash: How come? I was lumping them together under the idea that they were all testing different edge cases of the same function.

eliascodes: I guess it's a preference thing, but IMO it's hard to read and maintain long test cases.

react-backend/shared/controllers/feedback.js
```diff
@@ -1,21 +1,24 @@
...
-const route = ('/',
-(req, res, next) => {
+module.exports = (req, res, next) => {
   const pgPool = req.app.get('client:psql');
   const standardJwtSecret = req.app.get('cfg').session.standard_jwt_secret;
   const { feedbackScore } = req.body;
   const { authorization: token } = req.headers;
 
   jwt.verify(token, standardJwtSecret, (err, decoded) => {
+    if (err) return next;
```
eliascodes: Forgot a bit

tests/shared/controllers/feedback.test.js
```diff
@@ -43,9 +43,16 @@ test('POST /api/cb/feedback', tape => {
         .post('/api/cb/feedback')
         .set('authorization', token)
         .send(badPayload)
-        .expect(500)
-        .expect('Content-Type', /text/)
-        .end(async err => {
+        .expect(400)
```
eliascodes: Looking a bit closer at this test, the reason the request fails is not because the payload is bad. Actually there's nothing wrong with the payload at all. It's that the email in the token is non-existent. 

This means that the request is well formed, but unauthorized. That's a `401`, and the router should check whether the email exists first and respond with an appropriate message.

Why aren't you using the authentication middleware on this route? It would do all this for you.

astroash: Ahhh yeah, I should be using middleware 🙈 . Do you think I need an error test case, as this was checking the if the email grabbed from the jwt was linked to a valid cb. If it's running through middleware I guess this can never happen.

react-backend/shared/controllers/feedback.js
```diff
@@ -0,0 +1,24 @@
...
+  jwt.verify(token, standardJwtSecret, (err, decoded) => {
+    if (err) return next;
+
+    const { email: cbEmail } = decoded;
+
+    cbAdmin
+      .insertFeedback(pgPool, { cbEmail, feedbackScore })
+      .then(data => res.send({ result: data }))
+      .catch(() =>
+        res
```
eliascodes: This `catch` will catch _any_ error occurring during the promise chain. It should be used for errors that we cannot anticipate and/or mitigate in advance (like programmer errors).

We _can_ anticipate payload validation errors. You're not doing any validation in this controller. Then you can correctly respond with a `400` for payload errors, and leave this as a `500`.

tests/shared/models/cb-admin.test.js
```diff
@@ -0,0 +1,142 @@
+const test = require('tape');
+const pg = require('pg');
+const { getConfig } = require('../../../config');
+const { refresh: refreshDB } = require('../../../db/scripts');
+const cbAdmin = require('../../../react-backend/shared/models/cb-admin');
+
+const config = getConfig(process.env.NODE_ENV);
+
+test('Db Query | cbAdmin ', async tape => {
```
eliascodes: `CbAdmin model` is a better test suite descriptor.

tests/shared/models/cb-admin.test.js
```diff
@@ -0,0 +1,142 @@
...
+  const client = new pg.Client(config.psql);
+  await client.connect();
+
+  tape.test('cbAdmin.getFeedback | existing cb_id', async t => {
```
eliascodes: Since they're nested inside the `CbAdmin model` test suite, there's no need to prefix the method names with `CbAdmin`, since that is implicit.

eliascodes:
@astroash please rebase your branch onto master.

And btw, you need to leave a blank line between the commit summary and the long-form commit message for it to be recognised properly by all git clients:

screen shot 2018-03-14 at 14 25 07

react-backend/shared/controllers/feedback.js
```diff
@@ -0,0 +1,24 @@
+const jwt = require('jsonwebtoken');
+const cbAdmin = require('../models/cb-admin');
+
+module.exports = (req, res, next) => {
+  const pgPool = req.app.get('client:psql');
+  const standardJwtSecret = req.app.get('cfg').session.standard_jwt_secret;
+  const { feedbackScore } = req.body;
+  const { authorization: token } = req.headers;
+
+  jwt.verify(token, standardJwtSecret, (err, decoded) => {
```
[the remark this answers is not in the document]
astroash: It's the only way I could get info about which cb is logged in on the front end.

react-backend/client/src/components/visitors/thank_you_feedback.js
```diff
@@ -0,0 +1,32 @@
+import React, { Component } from 'react';
```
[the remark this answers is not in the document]
astroash: Sweet, that makes sense, should I get it to use [`this.props.history.goBack()`](https://reacttraining.com/react-router/web/api/history) instead?

react-backend/client/src/assets/icons/admin/Delete.svg
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
astroash: Cool, I've never heard of this before. Do I simply just run it though [this lovely site I found on google?](http://petercollingridge.appspot.com/svg-optimiser)

docs/api_migration_plan.md
```diff
@@ -82,3 +82,4 @@ All non-successful responses should return appropriate status codes in the range
 | `POST   /cbs/login` | | `{ query: { email, password } }` | `{ result: { cb_admin_token } }` | `POST /admin/login` |
 | `POST   /cbs/:id/emails` | | `{ query: { password_reset: Bool } }` | `{ result: null }` | `POST /cb/pwd/reset` |
 | `POST   /visitors/statistics` | `Authorization: cb_admin_token` | `{ query: { group_by: { gender, activity, age: [18, 35, 51, 70], days: ['monday'] } }` | `{ result: { gender: { male, female, ... }, activity: { yoga: 12, ...}, age: { 0-17: 2, ... }, days: { monday: 2, ... } } }` | `POST /users/chart-all` <br/> `POST /users/filtered` |
+| `POST   /cb/feedback` | `Authorization: cb_admin_token` | ` { email }` | `[ { cbEmail, feedbackScore, feedback_date } ]`| ✨ shiny shiny new ✨ |
```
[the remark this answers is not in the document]
astroash: ewwww conform...

react-backend/shared/controllers/feedback.js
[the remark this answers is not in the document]
eliascodes: Did you read the auth middleware?

react-backend/client/src/components/visitors/thank_you_feedback.js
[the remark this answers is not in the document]
eliascodes: Sure

react-backend/client/src/assets/icons/admin/Delete.svg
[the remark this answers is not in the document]
eliascodes: yeah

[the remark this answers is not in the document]
astroash: Are these default settings good?

screen shot 2018-03-14 at 18 31 13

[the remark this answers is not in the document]
eliascodes: No idea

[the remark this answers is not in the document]
astroash: LGTM :shipit:

tests/shared/models/cb-admin.test.js
astroash: I've named all my model tests like this to make it easier to read the tape output with the nested tests we're using

tests/shared/models/cb-admin.test.js
astroash: 🤷🏽‍♀️ Dunno why but I find this easier to read/reason about. Would prefer to leave it as is if that's okay.

react-backend/client/src/components/visitors/thank_you_feedback.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+export default () => (
```
eliascodes: Lol, these thanks components are still almost the same, but I'll let it go at this stage.

react-backend/client/src/components/hoc/redirect_after_timeout.js
```diff
@@ -0,0 +1,27 @@
+import React, { Component } from 'react';
+import { withRouter } from 'react-router-dom';
+import PropTypes from 'prop-types';
+
+const redirectAfterTimeout = (url, ms) => (Child) => {
```
eliascodes: This is really good.

docs/api_migration_plan.md
```diff
@@ -82,3 +82,4 @@ All non-successful responses should return appropriate status codes in the range
 | `POST   /cbs/login` | | `{ query: { email, password } }` | `{ result: { cb_admin_token } }` | `POST /admin/login` |
 | `POST   /cbs/:id/emails` | | `{ query: { password_reset: Bool } }` | `{ result: null }` | `POST /cb/pwd/reset` |
 | `POST   /visitors/statistics` | `Authorization: cb_admin_token` | `{ query: { group_by: { gender, activity, age: [18, 35, 51, 70], days: ['monday'] } }` | `{ result: { gender: { male, female, ... }, activity: { yoga: 12, ...}, age: { 0-17: 2, ... }, days: { monday: 2, ... } } }` | `POST /users/chart-all` <br/> `POST /users/filtered` |
+| `POST   /cb/feedback` | `Authorization: cb_admin_token` | ` { query: { feedbackScore }` | `[ { result: { cbEmail, feedbackScore, feedback_date } } ]`| ✨ shiny shiny new ✨ |
```
eliascodes: This is still inaccurate:

screen shot 2018-03-15 at 10 57 00

react-backend/client/src/components/visitors/thank_you_feedback.js
astroash: I was contemplating passing a key to `props` for the component to match on to add the correct copy, but I felt like this wasn't simplifying the already simplified functional component much and would take away from readability, so settled for this 🤷🏽‍♀️. I mean, if it was elm I would have union typed the crap out of it 😏. Happy to implement the above, or any other suggestion if ya want.

eliascodes: Nah leave it, just fix the docs.

PR #265, opened 2018-03-08
Api client

### LINKED ISSUE #264: Create HTTP API interface for React app
The API client acts as an interface between the HTTP API implemented by the express app and the client side React app. All communication between the two should go through this module.

The purpose of the API client is to abstract away the details of the HTTP API and instead present a simple and consistent interface for the React app to consume.

This will make changes to the HTTP API easier to accommodate, since it will usually only require changes to a single module, instead of several components.

eliascodes:
Fixes #264 

So this is where I'm at so far, the API client itself is basically done, with the following tasks remaining:
* [ ] Add tests
* [x] Substitute calls to `fetch(URL)` with calls to the client
* [x] Remove existing HTTP API helpers
* [ ] _(Extra credit)_ add response processing into the client to align the data the react app receives more closely with the [new API design](https://github.com/TwinePlatform/DataPower/blob/master/docs/api_migration_plan.md)

The way I've written the client, I've tried to make the API it exposes to the React app consistent, as far as is reasonable, with the new API design, converting inputs to be compatible with the existing API internally. This is why the `Visitors.get` method is relatively complex. All that logic can be deleted once `GET /visitors` works as specified in the new design.

### EDIT

OK "done" now. I've not added tests since they'd all have to be re-written after the API is migrated towards the new design. I've tested all the user flows I can think of and squashed any bugs arising from my changes.

I've not added any response processing because it would have inflated the diff considerably, and it can wait for a subsequent PR. Similarly I've removed some of the input processing (including the infamous `Infinity`...) for the same reason: it was just adding code that would soon be removed.


react-backend/client/api_client.js
```diff
@@ -0,0 +1,241 @@
...
+ */
+import axios from 'axios';
+import { zipWith } from 'ramda';
+
+const parseAgeFilter = (ageBrackets) => {
+  const agesLower = (ageBrackets[0] !== 0 ? [0] : []).concat(ageBrackets);
+  const agesUpper = agesLower
+    .map(x => x - 1)
+    .filter(x => x > 0)
+    .concat(Infinity);
```
astroash: This is an unnecessary use of `Infinity`, you can improve readability of this by using:
```
const parseAgeFilter = (ageBrackets) => {
  const agesLower = (ageBrackets[0] !== 0 ? [0] : []).concat(ageBrackets);

  const agesUpper = agesLower
    .map(x => x - 1)
    .filter(x => x > 0)
    .concat("false");

  return zipWith((x, y) => (y!=="false" ? `${x}-${y}` : `${x}-more`), agesLower, agesUpper)[0];
};
```

eliascodes: Disagree that this is more readable. Leaving aside the use of a string value that reads like a boolean, `"false"` is not an age, and so is out of place in `agesUpper`.

I may have agreed had you suggested something like:
```js
const agesUpper = agesLower
  .map(x => x - 1)
  .filter(x => x > 0)
  .concat('more');

return zipWith((x, y) => `${x}-${y}`, agesLower, agesUpper)
```

Because it allows us to remove the branch from the final line. Although note that this is mixing presentational concerns with data transformations; the `'more'` is how we represent an undefined upper bound using text.

Maybe you would prefer:
```js
const ageToStr = n => Number.isFinite(n) ? `${n}` : 'more'

const agesLower = (ageBrackets[0] !== 0 ? [0] : [])
  .concat(ageBrackets)
  .map(ageToStr);

const agesUpper = agesLower
  .map(x => x - 1)
  .filter(x => x > 0)
  .concat(Infinity)
  .map(ageToStr);

return zipWith((x, y) => (`${x}-${y}`), agesLower, agesUpper)[0];
```

astroash: > Disagree that this is more readable. Leaving aside the use of a string value that reads like a boolean, "false" is not an age, and so is out of place in agesUpper.

But following this logic, `Infinity` is also not an age.

How about:

```erlang
const parseAgeFilter = (ageBrackets) => {
  const agesLower = (ageBrackets[0] !== 0 ? [0] : []).concat(ageBrackets);
  
  const agesUpper = agesLower
    .map(x => x - 1)
    .filter(x => x > 0)
    .concat("okImSorryIllTryToStopMakingJokingInPRsMaybeINeedASignOrSomethingButITHoughtAStringifiedFalseWasEnough");

  return zipWith((x, y) => (typeof y !== 'String' ? `${x}-${y}` : `${x}-more`), agesLower, agesUpper)[0];
};
```

astroash: But on a serious note, what's the expected params/output for this function? I feel like I can get similar functionality from:
```js
const parseAgeFilter = (ageBrackets) => 
  ageBrackets[0]<= 1 ? `0 - ${ageBrackets[1]-1}` : `0 - ${ageBrackets[0]-1}`
```

eliascodes: I will eventually remove the `[0]` at the end of the return statement when we make changes to the API. Right now, the endpoint expects only one age-group (I believe), but we can make it more general.

eliascodes: > But following this logic, Infinity is also not an age.

Let's say an "age" is any finite positive integer (a natural number). What we are concerned with is describing a set of intervals over the whole natural number line. All but the last interval is trivial. So we are looking a semantic to represent the final interval which might loosely be described as "anything greater than N, inclusive". The appropriate semantic for this in mathematics is a half-open interval `[N, Inf)`. So `Infinity`, while not strictly speaking an age, is the appropriate value to use here.

astroash: **Disclaimer:** I'm not suggesting any of the code should be changed _but..._

... with the majority of time in dev world being spent on maintaining code (thus making readability financially beneficial), and the fair assumption that the majority of web developers do not have maths degrees, I'm merely pointing out that this is an "unusual" use of `Infinity` through poorly placed humour.

react-backend/client/src/components/admin/adminUserDetailsPage.js
```diff
@@ -83,28 +83,31 @@ export default class AdminUserDetailsPage extends Component {
...
 
-    adminPost(this, '/api/user/details/update', {
-      userId: this.state.userId,
-      userFullName: this.state.userFullName,
-      sex: this.state.sex,
-      yearOfBirth: this.state.yearOfBirth,
+    Visitors.update(this.props.auth, {
+      id: this.state.userId,
+      name: this.state.userFullName,
+      gender: this.state.sex,
```
astroash: Lol

tests/routes/visitor_check.test.js
```diff
@@ -21,7 +21,7 @@ test('POST /api/visit/check | viable name & email & exists in database', t => {
     .post('/api/visit/check')
     .set('authorization', token)
     .send(successPayload)
-    .expect(200)
+    .expect(409)
```
astroash: Is this test correctly named? It is meant to be a success case but returns a `409`.

tests/routes/visitor_check.test.js
```diff
@@ -45,7 +45,7 @@ test('POST /api/visit/check | viable name & email & not found in database', t =>
     .post('/api/visit/check')
     .set('authorization', token)
     .send(successPayload)
-    .expect(400)
+    .expect(200)
```
astroash: `200` feels a bit wrong here..

docs/architecture.md
```diff
@@ -67,3 +65,14 @@ This documents the desired architecture for this project. This was initially dis
 |  |- ...
 |- index.js
 ```
+
+### Tests
+Adhere to the `jest` convention of having a `__tests__` directory colocated with all files under-test. For example:
```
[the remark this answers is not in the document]
eliascodes: Location of tests isn't really an architectural consideration. This is really more of an organisational convention. I've put a general example here, we can work out how that relates to a specific file structure.

react-backend/client/src/api/index.js
```diff
@@ -0,0 +1,239 @@
+/*
+ *
+ */
+import axios from 'axios';
+
+
+export const Activities = {
+  get: (tkn, { weekday = 'all' } = {}) =>
```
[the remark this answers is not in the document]
eliascodes: Axios provides a much higher level API than `fetch`. Checkout the docs.

react-backend/client/src/api/index.js
```diff
@@ -0,0 +1,239 @@
...
+      {
+        org_name: orgName,
+        genre: category,
+        email,
+        uploadedFileCloudinaryUrl: logoUrl,
+      },
+      { headers: { Authorization: tkn } },
+    ),
+
+  delete: () => { },
```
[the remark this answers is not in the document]
eliascodes: Yes

react-backend/client/src/components/admin/adminUserDetailsPage.js
eliascodes: I know, I would have changed the state field too, but wanted to keep the diff small.

tests/routes/visitor_check.test.js
eliascodes: Where does it say it's meant to be a success case?

tests/routes/visitor_check.test.js
eliascodes: Why? The purpose of the endpoint is to validate whether the payload is valid for creating a new visitor record. If no existing visitor record with that email exists, it is valid.

eliascodes:
@RogeredBacon if you would prefer that, it's fine with me. Though it will result in a slightly larger diff in the end after incorporating the feedback stuff.

tests/routes/visitor_check.test.js
astroash: Ahhh I didn't notice you'd swapped the ternary in the route. This test makes more sense now.

eliascodes:
@astroash @RogeredBacon ready to merge AFAIK

Issue #210, opened 2018-02-23
DB schema review


eliascodes:
In the process of doing #209, I noticed a few things about the schema that should probably be changed.

* [ ] Many column names are in camel case. The widespread convention is for SQL column names to be in `snake_case`.
* [ ] The use of `date` as a column name, aside from being problematic for using a SQL keyword, is ambiguous. What date is being recorded here? The date when the row is created? Updated? A better name is necessary so it's purpose can be gleaned from reading the schema. It's common to have one field for the time a record was created and another for the time a record was last updated.
* [ ] Names should not be abbreviated (e.g. `hash_pwd` can simply be `password`, `cb_id` -> `community_business_id`, `cbusiness` -> `community_businesses`)
* [ ] `token` is ambiguous: what kind of token and what is it for? I think this is used for password resets, right?
* [ ] `tokenExpire` is a `BIGINT`, presumably it's meant to be a UNIX timestamp? In which case wouldn't `TIMESTAMP` be a more meaningful datatype?
* [ ] Is there are reason we are collecting the sex of users? Are we actually talking about sex, or are we talking about gender? Do we really need either?

No action yet, would prefer tests on the query functions first.

RogeredBacon:
The token expire is currently milliseconds until the token expires, which I think is 30mins, though maybe a better type of int could be used.

We're talking about sex rather than gender and this was a definite datapoint that Sonja wanted.

The real point of the app is to help CB's compete with larger companies in terms of tailoring their services towards certain customer types. The CB's need to know if they're only attracting young/old woman/men, so that they can then focus on their core users or try to diversify; and then whether what they're doing is working.

We've never really looked properly at the schema and I agree it definitely needs to be done, renaming the columns will make it a fair bit of work because of how much relies on them but the tests will hopefully sort most of it.

eliascodes:
These are all addressed by proposals in #357, so closing this.

PR #267, opened 2018-03-08
Signup Page

### LINKED ISSUE #257: As a user I can see a GDPR statement at the end of sign-up and have the opportunity to opt out of receiving emails/sms
As a visitor, I see a GDPR-compliant statement on marketing on the last page of sign up, and have the option to tick a box for receiving news from this CB by email:

- [ ] GDPR Statement page with check boxes for sms and email - *1 day*
- [ ] DB schema modifications and testing - *1.5 days*

### LINKED ISSUE #258: GDPR Statement page - See #257
- Modify existing page
- Add statement
- Add checkboxes with explanations

This is actually more work than initially suggested by the deliverables as currently this page is two pages. According to the redesigns we'll need to change the input fields, condense both pages together and *then* add the deliverables requirements.

Luckily we have scoped for implementing the redesign which this partly counts towards.

RogeredBacon:
I've changed the signup page to be a single form that is imported into main.js.
Added optional phone number input and contact check-boxes.

Fixed the tests so that they pass, though I haven't  touched the CSS yet.

The only errors I'm getting are about not having correct props validation in signup_form (Prop-Types). Not entirely sure how to deal with that properly...

Fixes #257 
Fixes #258


db/schema.sql
```diff
@@ -21,8 +21,11 @@ CREATE TABLE users
   sex VARCHAR(30) NOT NULL,
   yearOfBirth INTEGER NOT NULL,
   email VARCHAR(100),
+  phone VARCHAR(15),
   date TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
-  hash VARCHAR(64) NOT NULL
+  hash VARCHAR(64) NOT NULL,
+  emailContact BOOLEAN NOT NULL,
```
eliascodes: A common convention is to name Boolean fields like a yes/no question. Also as mentioned in #210, database columns should not be abbreviated, and we shouldn't worry too much about using long column names if they're more descriptive.

I'd suggest `is_email_contact_consent_granted` and `is_sms_contact_consent_granted`.

react-backend/client/src/components/visitors/main.js
```diff
@@ -22,8 +21,11 @@ export default class Main extends Component {
     this.state = {
       fullname: '',
       email: '',
-      sex: 'male',
-      year: 1980,
+      phone: '',
+      gender: '',
+      year: '',
+      emailContact: true,
```
eliascodes: are you sure you want the default values to be `true`?

react-backend/client/src/components/visitors/signup_form.js
```diff
@@ -0,0 +1,71 @@
...
+export default () => (
+  <div>
+    <section className="Main col-9">
+      <h1>Please tell us about yourself</h1>
+      {error && (
+        <div className="ErrorText">
+          {error.map(el => <span key={el}>{el}</span>)}
+        </div>
+      )}
+      <form className="Signup" onChange={this.handleChange}>
```
eliascodes: This wont work, you need to pass down the methods as props. Look up stateless functional components and how to pass props to them.

react-backend/client/src/components/visitors/signup_form.js
```diff
@@ -0,0 +1,71 @@
...
+        <Input question="Phone Number (optional)" option="phone" />
+        <Select
+          question="Gender"
+          option="gender"
+          choices={['male', 'female', 'prefer not to say']}
+        />
+        <Select question="Year of Birth" option="year" choices={years} />
+      </form>
+      <button onClick={this.checkUserExists} className="Button">
+        {' '}
```
eliascodes: This is unnecessary, JSX allows you to use text literals, just use a space

react-backend/database/queries/user_register.js
```diff
@@ -1,15 +1,29 @@
 const insertUser = `
-  INSERT INTO users (cb_id, fullName, sex, yearOfBirth, email, hash)
-  VALUES ($1, $2, $3, $4, $5, $6)`;
+  INSERT INTO users (cb_id, fullName, sex, phone, yearOfBirth, email, hash, emailContact, smsContact)
+  VALUES ($1, $2, $3, $4, $5, $6, $7)`;
```
eliascodes: There's 9 values

react-backend/client/src/components/visitors/signup_form.js
```diff
@@ -0,0 +1,71 @@
...
+        Next{' '}
```
eliascodes: Is `Next` the right copy here? Surely this is the submit button?

react-backend/client/src/components/visitors/signup_form.js
```diff
@@ -0,0 +1,71 @@
...
+        <Input question="Email" option="email" />
+        <Input question="Phone Number (optional)" option="phone" />
+        <Select
+          question="Gender"
+          option="gender"
+          choices={['male', 'female', 'prefer not to say']}
+        />
+        <Select question="Year of Birth" option="year" choices={years} />
+      </form>
+      <button onClick={this.checkUserExists} className="Button">
```
eliascodes: If this is the submit button, it should be a child of `form`.

react-backend/client/src/components/visitors/signup_form.js
```diff
@@ -0,0 +1,71 @@
...
+        and services you have requested from us, and improve how we deliver
+        those.
+        <br />
+        <br />
+        However, from time to time we would like to contact you with details of
+        other offers we provide. If you consent to us contacting you by email,
+        please tick to agree:{' '}
+      </p>
+      <form className="EmailCheckbox" onChange={this.handleChange}>
+        <Input type="checkbox" option="emailContact" checked />
```
eliascodes: This isn't what I meant; I meant that the checkboxes and the other fields should all be part of the _same_ `form` element. You can nest any element you like inside a `form`, so you can put your `section` and `aside` within it.

react-backend/client/src/components/visitors/signup_form.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import Input from './input';
+import Select from './select';
+
+export default () => (
+  <div>
+    <section className="Main col-9">
+      <h1>Please tell us about yourself</h1>
+      {error && (
```
eliascodes: This is a `ReferenceError`

astroash:
☝️ silly linting error 🙈

RogeredBacon:
Ya that's the prop-type stuff, I'm gonna mess around with it to try to get dread lord linter happy.

react-backend/client/src/components/visitors/signup_form.js
```diff
@@ -0,0 +1,67 @@
...
+    <form className="SignupForm" onChange={props.handleChange}>
+      <section className="Main col-9">
+        <h1>Please tell us about yourself</h1>
+        {props.error && (
+          <div className="ErrorText">
+            {props.error.map(el => <span key={el}>{el}</span>)}
+          </div>
+        )}
+        <Input question="Name" option="fullname" />
+        <Input question="Email" option="email" />
```
eliascodes: Any reason we're not using the `required` attribute on non-optional form inputs?

react-backend/client/src/components/visitors/signup_form.js
```diff
@@ -0,0 +1,67 @@
...
+        </button>
+      </section>
+    </form>
+    <form className="EmailSMSCheckbox" onChange={props.handleChangeCheckbox}>
+      <aside className="Privacy col-3">
+        <h3 className="Privacy__title">
+          Why are we collecting this information?
+        </h3>
+        <p className="Privacy__text">
+          Here at XXXXXXXX, we take your privacy seriously: we will only use
```
eliascodes: What's the plan with this `XXXXX...`?

react-backend/client/src/components/visitors/signup_form.js
```diff
@@ -0,0 +1,67 @@
...
+        <Input question="Name" option="fullname" />
+        <Input question="Email" option="email" />
+        <Input question="Phone Number (optional)" option="phone" />
+        <Select
+          question="Gender"
+          option="gender"
+          choices={['', 'male', 'female', 'prefer not to say']}
+        />
+        <Select question="Year of Birth" option="year" choices={props.years} />
+        <button onClick={props.checkUserExists} className="Button">
```
eliascodes: this `button` should have `type='submit'`

react-backend/client/src/components/visitors/main.js
```diff
@@ -1,19 +1,18 @@
...
-import Select from './select';
-import Button from './button';
 import qrcodelogo from '../../qrcodelogo.png';
-import FormPrivacy from '../visitors/form_privacy';
-import FormPrivacy2 from '../visitors/form_privacy2';
+import SignupForm from '../visitors/signup_form';
 import NotFound from '../NotFound';
 import errorMessages from '../errors';
 
 const generateYearsArray = (startYear, currentYear) =>
```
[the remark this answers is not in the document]
RogeredBacon: We googled the oldest living person in the UK, just in case they visit

react-backend/client/src/components/visitors/signup_form.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import Input from './input';
+import Select from './select';
+
+export default props => (
+  <div>
+    <form className="SignupForm" onChange={props.handleChange}>
```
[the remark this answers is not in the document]
RogeredBacon: That was how I was doing it originally, but I needed to write a slightly different handChange function for the checkboxes. Can I combine them and still use two different onChange functions?

react-backend/client/src/components/visitors/signup_form.js
RogeredBacon: oops completely forgot about that, that's meant to be the cb's name

eliascodes: You just need to differentiate between checkboxes and other input types in your change handler. I think you can inspect `e.target.type`.

RogeredBacon: I'm gonna leave the [insert cb name] bit for a seperate pr as it will potentially involve a new query/route.

react-backend/client/src/components/visitors/signup_form.js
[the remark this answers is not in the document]
eliascodes: Oops, I posted this below by accident, but it was meant for here:

> You just need to differentiate between checkboxes and other input types in your change handler. I think you can inspect `e.target.type`.

react-backend/client/src/components/visitors/signup_form.js
eliascodes: That's fine, please create an issue for it

react-backend/client/src/components/visitors/signup_form.js
```diff
@@ -12,21 +13,20 @@ export default props => (
             {props.error.map(el => <span key={el}>{el}</span>)}
           </div>
         )}
-        <Input question="Name" option="fullname" />
-        <Input question="Email" option="email" />
+        <Input question="Name" option="fullname" required />
```
eliascodes: AFAICS, this won't have any impact on the actual `input` element inside the custom `Input` component, because additional props (in this case `required`) are not passed through to the child `input` element.

RogeredBacon: What would be the right way of doing this? Would I need to edit the actual input.js component?

eliascodes:
```js
/* SignupForm.js */
// ...
<Input question="Name" option="fullname" required />
// ...


/* Input.js */
// ...
render () {
  // ...
    <input id={label} {...this.props} />
  // ...
}
// ...
```

eliascodes: Lookup "spreading props"

RogeredBacon:
Ok hopefully everything should be good now :crossed_fingers:

react-backend/client/src/components/visitors/input.js
```diff
@@ -25,6 +25,7 @@ export default class Input extends Component {
           onChange={this.handleUserInput}
           value={this.state.userInput}
           className="Form__Input"
+          {...this.props}
```
eliascodes: A safer pattern is to separate the props that you know you're going to use manually from rest, which you just want to pass through. You can use object rest spreading to do this easily.

```js
const { label, question, name, option, ...props } = this.props;

<input id={label} name={name} {...props} />
```
This prevents you from overwriting props that you've set manually, like would probably happen in this case, with `name` being set to `undefined` if `props.name` is empty, even if `props.option` is defined.

This happens because in the child component react does a shallow merge of the props
```
{ id: label, name: name, ...props }
```
And therefore any manually specified prop with the a key that exists in `props` would be overwritten.

react-backend/client/src/components/visitors/main.js
```diff
@@ -35,6 +35,7 @@ export default class Main extends Component {
   }
 
   handleChange = (e) => {
+    console.log(e.target.type);
```
eliascodes: Remove please

react-backend/client/src/components/visitors/input.js
RogeredBacon: Alrighty, I think I've done it

RogeredBacon:
Should I rebase this? I'm not sure its needed...

eliascodes:
@RogeredBacon There's an inconsistency between your changes to the API documentation and the codebase.

In the docs:
```
| `POST /user/details/update` | `Authorization: cb_admin_token` | `{ userId, userFullName, sex, yearOfBirth, email, phone, is_email_contact_consent_granted, is_sms_contact_consent_granted }` | `{ details: { `☝️` } }` |
```

In the code: 
https://github.com/TwinePlatform/DataPower/blob/f579387f14f57b8e13dd62eb67e962303a87fbc0/react-backend/routes/user_details_update.js#L5-L13

Could you please throw up a quick PR to fix this? Thanks.

Issue #253, opened 2018-03-06
Client side code review


eliascodes:
Some observations and comments:

#### General
* [ ] `<br/>` tags are an anti-pattern. Style the text properly with block-styling.

#### Auth components:
* [x] We're using the `fetch` API directly. `fetch` is a low-level interface that requires a lot of configuration that can be abstracted by a higher level library. We can either use something like `axios` or `superagent`, or write our own wrapper.
* [x] We should also abstract away the precise endpoints we're using into a wrapper. This has two advantages: first, if/when the endpoints change, we only need to change them in one place on the client. Second, using (sensible) method names instead of URLs is more readable and declarative.
* [x] Mixing error handling with success code paths (this is a side-effect of using `fetch` directly)
* [x] Repeated patterns not-exploited:
   * submit handler making an API call and then setting state
   * identical change handler

#### Errors
* [x] Error page can just be one paramterised component
* [ ] Error copy should be plain-text, injected by the component

#### Visitors
* [x] `Input` doesn't need internal state.
* [x] `Logoutbutton` does too much, should just be presentational component, decisions about `localStorage` and the like should reside in higher level components.
* [ ] `main.js` is too vague a filename
* [ ] Fetching data from `localStorage` directly in a class property declaration instead of in a lifecycle method
* [ ] Another anti-pattern: `throw`-ing an error from a `catch` (qrcode.js)

astroash:
@eliascodes are you happy for this to be closed? If not it's worth moving any remaining problems their own issue.

eliascodes:
Not yet, some of these are still problems, IMO.

Issue #132, opened 2018-01-07
Joi validation on the backend


finnhodgkin:
I'm a big fan of [joi](https://github.com/hapijs/joi) validation ([plus express module](https://www.npmjs.com/package/express-joi-validator)) and it's pretty easy to hook up to express. It would allow all validation to be kept in one format and also remove bloat from some of the `post` routes.

Error handling is kept in separate middleware so the validation errors could all be in one place too 🎉. 

```js
const visitorSchema = {
  body: {
    'Full name': joi
      .string()
      .required()
      .min(1),
    Email: joi
      .string()
      .required()
      .email(),
  },
};

router.post('/', expressJoi(visitorSchema), async (req, res) => {
  try {
    const { 'Full name': name, Email: email } = req.body;
    const { rows: [{ exists }] } = await getUserAlreadyExists(
      name.toLowerCase(),
      email,
    );
...
```

RogeredBacon:
Pretty sure this is no longer relevant so I'm closing this

astroash:
@eliascodes are you happy for this to be closed?

eliascodes:
Sure

PR #277, opened 2018-03-14
Joi validation

eliascodes:
Related to #132 and #253 

~~Branched off `api-client` in #265 so diff is bigger than in reality. Will rebase after it's merged.~~

Added two middleware to simplify request validation with `Joi` schema, which IMO is nicer and more powerful than `validator`.

Will add an example shortly, which will show both how to perform validation on the backend, and how to interpret the error messages on the front-end.


astroash:
Ohhh sorry, my bad, I didn't realise @RogeredBacon hadn't reviewed 😔

PR #278, opened 2018-03-14
Actually not so quick fix for props passed to user_details_update route

RogeredBacon:
Changed adminUserDetailsPage and its routes to be able to change a users phone number, email contact consent and sms contact consent.

Also had to change db query to stop it returning snake_case key names as dread lord linter had a bad experience with snakes and much prefers camels.

:snake: :camel: :snake: :camel: :snake: :camel: :snake: :camel: :snake: :camel: :snake: :camel: :snake: :camel: :snake: :camel: :snake: :camel: :snake: :camel: :snake: :camel: :snake: :camel: :snake: :camel: :snake: :camel: :snake: :camel: :snake:

### DIFF
--- a/react-backend/client/src/components/admin/adminUserDetailsPage.js
+++ b/react-backend/client/src/components/admin/adminUserDetailsPage.js
@@ -22,6 +22,9 @@ export default class AdminUserDetailsPage extends Component {
       url: '',
       errorMessage: '',
       cb_logo: '',
+      phoneNumber: '',
+      emailContact: false,
+      smsContact: false,
     };
   }
 
@@ -46,7 +49,17 @@ export default class AdminUserDetailsPage extends Component {
       });
   }
 
-  setUser = ({ fullname, sex, yearofbirth, email, date, hash }) => {
+  setUser = ({
+    fullname,
+    sex,
+    yearofbirth,
+    email,
+    phone_number,
+    date,
+    hash,
+    is_email_contact_consent_granted,
+    is_sms_contact_consent_granted,
+  }) => {
     this.setState({
       userFullName: fullname,
       sex,
@@ -54,6 +67,9 @@ export default class AdminUserDetailsPage extends Component {
       email,
       signupDate: date.replace(/T/g, ' ').slice(0, 19),
       hash,
+      phoneNumber: phone_number,
+      emailContact: is_email_contact_consent_granted,
+      smsContact: is_sms_contact_consent_granted,
       errorMessage: '',
     });
   };
@@ -64,12 +80,24 @@ export default class AdminUserDetailsPage extends Component {
     });
   };
 
-  handleChange = e =>
-    this.setState({
-      [e.target.name]: e.target.value,
-      errorMessage: '',
-      successMessage: '',
-    });
+  handleChange = (e) => {
+    switch (e.target.type) {
+      case 'checkbox':
+        this.setState({
+          [e.target.name]: e.target.checked,
+          errorMessage: '',
+          successMessage: '',
+        });
+        break;
+      default:
+        this.setState({
+          [e.target.name]: e.target.value,
+          errorMessage: '',
+          successMessage: '',
+        });
+        break;
+    }
+  };
 
   handleChangeSex = e => this.setState({ sex: e.target.value });
 
@@ -89,6 +117,9 @@ export default class AdminUserDetailsPage extends Component {
       sex: this.state.sex,
       yearOfBirth: this.state.yearOfBirth,
       email: this.state.email,
+      phoneNumber: this.state.phoneNumber,
+      emailContact: this.state.emailContact,
+      smsContact: this.state.smsContact,
     })
       .then(res => res.details)
       .then(this.setUser)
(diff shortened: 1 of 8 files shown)


RogeredBacon:
I'm having an issue with lint wanting a variable name to be camel case where that name is defined by the db column. On line 57, 60, 61 in adminUserDetailsPage.js

react-backend/client/src/components/admin/adminUserDetailsPage.js
```diff
@@ -46,14 +49,27 @@ export default class AdminUserDetailsPage extends Component {
...
+  setUser = ({
+    fullname,
+    sex,
+    yearofbirth,
+    email,
+    phone_number,
+    date,
+    hash,
+    is_email_contact_consent_granted,
+    is_sms_contact_consent_granted,
```
RogeredBacon: Linter error here with the db variables not being in camel case

RogeredBacon:
Alrighty everything should be working now. Probably for the best the query is more specific now, rather than getting *EVERYTHING* from the row.

eliascodes:
@RogeredBacon seems to be a bug:
1. Go user details page for a user with "yes" for both contact fields
2. Un-tick one of the contact fields.
3. Press submit
4. You will see everything update as expected
5. Re-tick the field you un-ticked
6. Press submit
7. Expect both fields to be ticked, but both become un-ticked.

RogeredBacon:
Ok, it should work now

react-backend/routes/user_details_update.js
```diff
@@ -9,7 +9,10 @@ router.post('/', (req, res, next) => {
     req.body.userFullName,
     req.body.sex,
     req.body.yearOfBirth,
-    req.body.email
+    req.body.email,
+    req.body.phoneNumber,
+    req.body.emailContact,
+    req.body.smsContact
```
eliascodes: You've no validation in this route, which means `emailContact` and `smsContact` could be `undefined`. But the schema says:
```sql
is_email_contact_consent_granted BOOLEAN NOT NULL,
is_sms_contact_consent_granted BOOLEAN NOT NULL
```
So passing `undefined` will cause a SQL error. Better to have sensible defaults in the schema.

tests/database/queries/user_details.test.js
```diff
@@ -23,12 +23,12 @@ test('DB Query | user_details', async tape => {
...
           yearofbirth: 1984,
           email: 'hello@yahoo.com',
-          phone_number: '7538654284',
+          phone: '7538654284',
           date: new Date('Mon May 15 2017 12:24:57 GMT+0000 (UTC)'),
           hash:
             '9fb59d630d2fb12f7478c56c5f1b2fff20e0dd7c9d3a260eee7308a8eb6cd955',
-            is_email_contact_consent_granted: true,
-            is_sms_contact_consent_granted: true,
+            emailcontact: true,
```
eliascodes: The indentation here is wrong and misleading, should be inline with `hash`.

tests/database/queries/user_details_update.test.js
```diff
@@ -33,12 +33,12 @@ test('DB Query | user_details_update', async tape => {
...
         yearofbirth: 2001,
         email: 'not@makingsense.com',
-        phone_number: '07534532459',
+        phone: '07534532459',
         date: new Date('Mon May 15 2017 12:24:57 GMT+0000 (UTC)'),
         hash:
           '9fb59d630d2fb12f7478c56c5f1b2fff20e0dd7c9d3a260eee7308a8eb6cd955',
-          is_email_contact_consent_granted: true,
-          is_sms_contact_consent_granted: true,
+          emailcontact: true,
```
eliascodes: should add a test case where `emailcontact` and `smscontact` are undefined, since that is possible.

react-backend/routes/user_details_update.js
RogeredBacon: So do you mean removing the `NOT NULL` and adding something like `DEFAULT false`? Or changing the route so that it email/smsContact cannot be undefined?

eliascodes: > So do you mean removing the NOT NULL and adding something like DEFAULT false?

I mean `column BOOLEAN NOT NULL DEFAULT false`

> Or changing the route so that it email/smsContact cannot be undefined?

We will eventually add validation to this endpoint, but after my changes are merged.

> Would the db Schema default not be irrelevant for the update route as the row will already exist?

Your query text specifies placeholders for both the consent fields, and if you pass `undefined` to `pg` for those placeholders, it will try to insert `NULL` values in those fields, which is not allowed.

RogeredBacon: I'm not sure I understand you, if I change the schema to `column BOOLEAN NOT NULL DEFAULT false` and then try to pass it undefined for the consent values, it still errors as its breaking the  `NOT NULL` constraint.

If I make the schema change and then run a test on the route without passing any value for the consent columns such as:
`` const actual = await putNewUserDetails(
        client,
       1,
      1,
     'Little Nonsense',
    'female',
   2001,
  'not@makingsense.com',
 '07534532459'
      );
``
I get an error telling me that the `NOT NULL` constraint has been violated.

RogeredBacon: So in terms of this pr should I add the validation to change the query depending on if the consents have been defined? It seems like if we intend to add validation to the route anyway it would be quickly deprecated?

eliascodes: The easiest way around this now is to add default values of `false` to the function arguments in `user_details_update.js`.

eliascodes:
@RogeredBacon I wanted to get this in, so I made the final fix myself. See the commit message on aa5ef71 for details of what I meant in my last review.

Thanks.

PR #281, opened 2018-03-18
Feedback cb data vis

### LINKED ISSUE #280: View for cb-admin to see their feedback
Look into using chart.js 2 to render pretty little graphs about their feedback

- [x] have date input boxes for them to select a date range
- [x] add route to make query from db (route already written)
- [x] look [kinda like this](https://projects.invisionapp.com/share/N5G5B0RGRY7#/screens/282728496), but not really

astroash:
~~**#262 MUST BE MERGED FIRST**~~
~~Set db-feedback as base to see changes, will change to master when ready to merge~~

~~**This isn't styled but I'm keen to get it reviewed, even if it's not merged**~~

- adds axios query to get cb feedback
- update route to grab info from query string rather than body
- add HOC for cb admin auth check 
- component for feedback view
- shared button component using `styled-components` button
- update feeback query to return count
- fix all the test that above mucked up
- snuck in a script to start client from root

Relates #255 
Fixes #280


eliascodes:
@astroash #262 has been merged so you can rebase this onto master.

astroash:
@eliascodes, do you think I should refactor other `Components`, using the same admin auth check pattern in this PR?

react-backend/shared/controllers/feedback.js
```diff
@@ -1,24 +1,48 @@
-const cbAdmin = require('../models/cb-admin');
+const cbAdmin = require('../models/cb_admin');
 
-module.exports = (req, res, next) => {
-  const pgPool = req.app.get('client:psql');
+module.exports = {
+  get: (req, res, next) => {
+    const pgPool = req.app.get('client:psql');
+    try {
```
eliascodes: You've wrapped this all in a try/catch. What exception are you expecting might be thrown?

react-backend/router.js
```diff
@@ -40,7 +40,8 @@ router.use('/cb/pwd/change', cbPasswordChange);
 router.use('/cb/pwd/reset', cbPasswordReset);
 
 // Not open but following api plan
-router.post('/cb/feedback', mwIsAuthenticated, cbFeedback);
+router.get('/cb/feedback', mwAdminIsAuthenticated, cbFeedback.get);
```
eliascodes: I would take this opportunity to start removing things from this router file. My intention was for the `controllers` directory in each "feature" directory to export a router object.

So that means `/cb-admin/controllers` exports a router containing all `/api/cb` routes.

react-backend/shared/controllers/feedback.js
```diff
@@ -1,24 +1,48 @@
-const cbAdmin = require('../models/cb-admin');
```
eliascodes: This shouldn't be in shared, as per above.

react-backend/shared/controllers/feedback.js
```diff
@@ -1,24 +1,48 @@
...
+const cbAdmin = require('../models/cb_admin');
 
-module.exports = (req, res, next) => {
-  const pgPool = req.app.get('client:psql');
+module.exports = {
+  get: (req, res, next) => {
```
eliascodes: You should use the joi validation middleware. I'd suggest one of the following:

1. instead of a function, `exports.get` and `exports.post` export an array which is the middleware stack for that route. This can then be spread into the call to `router.<METHOD>` e.g. `router.post('/route', othermiddleware, ...stack)`.
2. explicitly namespace the `exports.get` object using `handler`, `validation` (and possibly later on `auth`) keys.

react-backend/client/src/components/hoc/admin_login_check.js
```diff
@@ -0,0 +1,74 @@
...
+        this.cancelRequest = () => {
+          reject(new Error('REQUEST_CANCELLED'));
+        };
+        fetch('/api/admin/check', opts)
+          .then(resolve)
+          .catch(reject);
+      })
+        .then(() => this.setState({ auth: requestStates.SUCCESS }))
+        .catch((error) => {
+          if (error.message === 'REQUEST_CANCELLED') return;
```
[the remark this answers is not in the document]
astroash: https://www.ecma-international.org/ecma-262/6.0/#sec-promise-objects

react-backend/shared/controllers/feedback.js
astroash: Just incase since/until are not date worthy

react-backend/client/src/components/hoc/admin_login_check.js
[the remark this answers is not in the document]
eliascodes: OK, well I googled "check if component is still mounted before doing something react" and [this is the first result](https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html)

react-backend/shared/controllers/feedback.js
eliascodes: Passing values that can't be parsed by the `Date` constructor won't cause it to  throw. Instead the postgres client will cause the promise to reject, which is covered by `catch`

react-backend/client/src/components/hoc/admin_login_check.js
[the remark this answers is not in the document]
astroash: Thanks, this seems to have fairly similar functionality, but with mine cancelling immediately rather than waiting for the fetch to return. Although the article still doesn't answer the question of where to store the cancellable promise.

TBH I don't see the need to pick between either as there's the eagerness to move to redux in the long run.

react-backend/router.js
astroash: Can we move all files in [one big swoop?](https://github.com/TwinePlatform/DataPower/issues/296) I'm starting to have some sad rebasing woes.

react-backend/shared/controllers/feedback.js
astroash: Is there an example of this in the codebase at the moment? I'm a little lost

react-backend/client/src/components/admin/adminFeedbackPage.js
```diff
@@ -0,0 +1,158 @@
...
+
+  handleGetFeedback= () => {
+    CbAdmin.getFeedback(this.props.auth, this.state.startDate, this.state.endDate)
+      .then(({ data }) => {
+        data.result[0] //eslint-disable-line
+          ? this.setState({ data, error: null })
+          : this.setState({ error: 'Sorry, no data was found', data: null });
+      })
+      .catch((err) => {
+        console.log(err);
```
eliascodes: If this comes back as a `401` we should redirect to login

react-backend/client/src/components/visitors/button.js
```diff
@@ -1,11 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { PrimaryButtonNoFill } from '../../shared/components/form/base';
 
-const button = props => (
-  <button className="Button" type="submit">
-    {props.label}
-  </button>
-);
+const button = props => <PrimaryButtonNoFill type="submit">{props.label}</PrimaryButtonNoFill>;
```
eliascodes: Why not just use `PrimaryButtonNoFill` directly?

react-backend/router.js
eliascodes: These are new routes, they shouldn't cause any rebase issues. What problems are you having?

react-backend/shared/controllers/__tests__/feedback.test.js
```diff
@@ -6,6 +6,62 @@ const { getConfig } = require('../../../../config');
 
 const config = getConfig(process.env.NODE_ENV);
 
+test('GET /api/cb/feedback', tape => {
+  const app = createApp(config);
+  const dbConnection = app.get('client:psql');
+
+  tape.test('GET /api/cb/feedback | successful query', t => {
```
eliascodes: Missing test case for date-limited feeback

react-backend/shared/controllers/feedback.js
eliascodes: https://github.com/TwinePlatform/DataPower/blob/master/react-backend/routes/cbauthentication/cb_register.js

react-backend/shared/controllers/feedback.js
```diff
@@ -1,24 +1,55 @@
-const cbAdmin = require('../models/cb-admin');
+const cbAdmin = require('../models/cb_admin');
 
-module.exports = (req, res, next) => {
-  const pgPool = req.app.get('client:psql');
+const createDate = str => {
```
eliascodes: As I said in [my comment](https://github.com/TwinePlatform/DataPower/pull/281#discussion_r176271323), this is unnecessary.

astroash: I wanted to return a `400` if incorrect data was sent, so this was my attempt at type safing the query. In my current layout any postgres error is being caught and passed to middleware returning a `500`.

react-backend/client/src/api/index.js
```diff
@@ -244,6 +244,15 @@ export const CbAdmin = {
       },
     ),
 
+  getFeedback: (tkn, since, until) =>
+    axios.get(
+      '/api/cb/feedback',
+      { params: {
```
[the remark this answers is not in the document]
astroash: Axios adds [query strings on get requests using `params:`](https://github.com/axios/axios/blob/master/README.md#request-config) and [you can access them with express on `req.query`](http://expressjs.com/en/api.html#req.query). If it's sent under the key `query`, you would have to access it in express as `req.query.query` 😩

react-backend/client/src/components/admin/adminFeedbackPage.js
```diff
@@ -0,0 +1,158 @@
...
+  },
+  {
+    feedback_score: 1,
+    label: 'Impressed',
+    backgroundColor: '#FDBD2D',
+    hoverBackgroundColor: '#DE9B06',
+  },
+];
+
+const donofig = (colorConfig, data) =>
```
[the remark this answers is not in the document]
astroash: But it's the donut config... 🙄

react-backend/client/src/components/admin/adminFeedbackPage.js
```diff
@@ -0,0 +1,158 @@
...
+            </PrimaryButton>
+            <PrimaryButton>
+              <h3>Dates between...</h3>
+            </PrimaryButton>
+          </div>
+
+          <div className="daterange-picker">
+            <DateRangePicker
+              startDate={this.state.startDate}
+              startDateId="your_unique_start_date_id"
```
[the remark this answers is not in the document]
astroash: Airbnb said this it [the minimum REQUIRED setup.](https://github.com/airbnb/react-dates#daterangepicker) They used CAPS, I questioned nothing.

react-backend/shared/controllers/feedback.js
eliascodes: Good point. Would be caught by Joi if you use the validation middleware though :wink:

react-backend/client/src/api/index.js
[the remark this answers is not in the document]
eliascodes: I was more talking about using the `filter` field, since what you're doing is filtering feedback by date. No point having generic query fields if we don't use them.

react-backend/client/src/components/admin/adminFeedbackPage.js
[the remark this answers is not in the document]
eliascodes: What does it do?

react-backend/client/src/components/admin/adminFeedbackPage.js
```diff
@@ -0,0 +1,158 @@
...
+    };
+  }
+  componentDidMount() {
+    this.handleGetFeedback();
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+
+    if (
+      prevState.focusedInput !== this.state.focusedInput
```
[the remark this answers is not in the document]
astroash: Part of the reason I refactored these request checkers is to keep the code in one place rather than passing them in as callbacks to `setState`.

With the datepicker I am using the change in `focusedInput` like an event listener to kick off the request. From what I can gather from documentation (which isn't great) and gh issues, this is the expected use.
 

Date picker request checks:
```
      prevState.focusedInput !== this.state.focusedInput // checks that it is called by a change in date picker
      && !this.state.focusedInput // checks that datepicker is being closed rather than opened
      && this.state.startDate // checks startDate is set
      && this.state.endDate // checks endDate is set (as both are needed)
```
All request checks:
```
      prevState.lastCall === lastCallStates.DATERANGEPICKER
      && this.state.lastCall === lastCallStates.ALL
```
☝️  This check is avoiding you from making fetch requests by repeatedly clicking 'ALL'. It will only fire another all request after a date picker request has been made.

react-backend/client/src/components/admin/adminFeedbackPage.js
```diff
@@ -0,0 +1,158 @@
...
+        data: feedbackColors.map(
+          el => data.filter(dat => dat.feedback_score === el.feedback_score)[0].count,
+        ),
+        backgroundColor: feedbackColors.map(el => el.backgroundColor),
+        hoverBackgroundColor: feedbackColors.map(el => el.hoverBackgroundColor),
+      },
+    ],
+  })
+;
+const lastCallStates = {
```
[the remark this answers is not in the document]
astroash: This was added to be explicit and make ["all request checks"](https://github.com/TwinePlatform/DataPower/pull/281#discussion_r176947273) easier to read.

react-backend/shared/controllers/feedback.js
astroash: 🙈 it's on my to do list...

react-backend/client/src/components/admin/adminFeedbackPage.js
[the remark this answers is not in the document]
astroash: 🤷🏽‍♀️

react-backend/client/src/components/admin/adminFeedbackPage.js
```diff
@@ -0,0 +1,158 @@
...
+  },
+];
+
+const donofig = (colorConfig, data) =>
+  ({
+    labels: feedbackColors.map(el => el.label),
+    datasets: [
+      {
+        data: feedbackColors.map(
+          el => data.filter(dat => dat.feedback_score === el.feedback_score)[0].count,
```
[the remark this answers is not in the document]
astroash: Up for other suggestions on how to display data in a set order...

react-backend/router.js
astroash: I just keep loosing bit's n pieces 😕.

react-backend/client/src/components/admin/adminFeedbackPage.js
astroash: How come?

react-backend/client/src/components/visitors/button.js
astroash: Yeah, I didn't end up using this component in the end. What's the best way to revert changes from this commit, should I soft reset?

react-backend/client/src/components/admin/adminFeedbackPage.js
eliascodes: It's better UX.

react-backend/client/src/components/admin/adminFeedbackPage.js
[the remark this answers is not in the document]
eliascodes: You should use the [onClose](http://airbnb.io/react-dates/?selectedKind=DRP%20-%20Calendar%20Props&selectedStory=with%20onClose%20handler&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) handler instead of relying on focus.

react-backend/client/src/components/admin/adminFeedbackPage.js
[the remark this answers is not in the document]
eliascodes: Ignore

react-backend/client/src/api/index.js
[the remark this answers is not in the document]
astroash: > No point having generic query fields if we don't use them.

What do you mean by this?

react-backend/client/src/shared/components/form/base.js
```diff
@@ -89,6 +92,9 @@ export const SecondaryButton = Button.extend`
 export const PrimaryButtonNoFill = PrimaryButton.extend`
   background: transparent;
   border-color: ${colors.highlight_primary};
+  &:hover{
+    background:transparent
```
eliascodes: Is this necessary? It's also transparent in unhovered state.

astroash: Yes, it's extended from `PrimaryButton` which was adding a hover colour.

astroash:
> When is styling happening? In this PR or another one?

It's styled in this PR

eliascodes:
this is what I see when I checkout your branch

screen shot 2018-03-27 at 17 22 49

astroash:
@eliascodes lol, yeah my styling sucks...

The [designs](https://projects.invisionapp.com/share/N5G5B0RGRY7#/screens/282728496) aren't very accurate. What else do you think needs to be changed?

RogeredBacon:
Perhaps change impressed/unimpressed to dissatisfied/satisfied as the title is satisfaction not impression.

Otherwise 🤷‍♂️ , like you say, the designs are unhelpful. Other than introducing a brand component to the page; feedback thermometer (good/ok/bad) for the last week for example, I'm not sure there's anything else to add other than maybe changing your html to the styled-components ones `<div>` -> `<FlexContainerCol>` etc...

eliascodes:
Stacking "Back to main page" and "logout" on top of each other, with neither aligned with the heading, looks bad.

I would have put one on either side of the heading, personally.

astroash:
# It's _vaguely prettier_

HMU if you guys have any other suggestions:
screen shot 2018-03-27 at 9 36 33 pm

eliascodes:
Last thing: 

screen shot 2018-03-28 at 09 33 49

I don't think it looks good to have the date picker smooshed in the little gap between the other buttons and the graph, especially when there's so much space to the right of the buttons. The datepicker should also have the same font-size/line-height as the buttons for consistency.

RogeredBacon:
░░░░░░░░░░░░░░░░░▒█▒░░░░░░░░░░░░░░░
░░░░░░░░░░░░░░░░▒█▓▓▒░░░░░░░░░░░░░░
░░░░░░░░░░░░░░░░██▓▒▓░░░░░░░░░░░░░░
░░░░░░░░░░░░░░░░███▓▒▓░░░░░░░░░░░░░
░░░░░░░░░░░░░░░░██▓▓▓██░░░░░░░░░░░░
░░░░░░░░░░░░░░░░██▓▓▓███░░░░░░░░░░░
░░░░░░░░░░░░░░░░▒██▓▓▓███▒░░░░░░░░░
░░░░░░░░░░░░░░░░░██▓▓▓▓███▓░░░░░░░░
░░░░░░░░░░░░░░░░░░██▓▓▓▓████░░░░░░░
░░░░░░░░░░░░░░░░░░░██▓▓▓▓▓███▒░░░░░
░░░░░░░░░░░░░░░░░░░░██▓▓▓▓▓▓██▓░░░░
░░░░░░░░░░░░░░░░░░░░▒█▓▓▓▓▓▓▓███▒░░
░░░░░░░░░░░░░░░░░░░░░██▓▓▓▓▓▓▓▓▓██░
░░░░░░░░░░░░░░░░░░░░░██▓▓▓▓▓▓▓▓▓▓██
░░░░░░░░░░░░░░░░░░░░░████▓▓▓▓▓▓▓▓▓█
░░░░░░░░▒▒▒▒░░░░░▒▒▒▓█▓▓▓▓▓▓▓▓▓▓▓▓▓
░░░░░░▓██████▓▓▓▓▓██▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
░░░░░█████▓▒▒▒▒▒▒▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
░░░░▒██▓▓▓▒▒▒▒▒▒▒▒▒▒▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
░░░░░██▓▓▓▒▒▒▒▒▒▒▒▒▒▒▓▓▓▓▓▓▓▓▓▓▓▓▓▓
░░░░░░█▓▓▓▒▒▒▒▒▒▒▒▒▒▒▒▓▓▓▓▓▓▓▓▓▓▓▓▓
░░░░░░░███▒▒▒▒▒▒▒▒▒▒▒▒▒▓▓▓▓▓▓▓▓▓▓▓▓
░░░░░░██▓▓▒▒▒▒▒▒▒▒▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
░░░░░░██▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▓▒▓▓▓▓▓▓▓▓▓▓
░░░░░░██▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▓▓▓▓▓▓▓▓▓
░░░░░░██▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▓▓▓▓▓▓
░░░░░░░▓▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▓▒▓▓▓
░░░░░░░▓█▒▒▒▒▒▒▒▒▓▒▓▒▓▒▒▒▒▒▒▒▒▒▓▒▓▓
░░░░░░░█▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▓▓
░░░░░░░█▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▓
░░░░░░░▓█▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
░░░░░░░░▓█▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
░░░░░░░░░▒█▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▓
░░░░░░░░░░█▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▓▓▓
░░░░░░░░░░██▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▓▓▓▓▓▓▓
░░░░░░░░░░▒█▓▒▒▒▒▒▒▒▒▒▒▒▒▓▓▓▓▓█████
░░░░░░░░░░░▒█▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▒░░

PR #288, opened 2018-03-20
React-Chartjs-2 Upgrade

### LINKED ISSUE #275: Upgrading our data vis module
I propose that we swap our current react-chartjs module for react-chartjs-2.

Basically it's an improvement on our current module in [every way](https://www.npmjs.com/package/react-chartjs-2); we wanted to do this originally but chartjs 2 had just been released and their wasn't a good react wrapper for it yet.

This will break all of our visualisations, but shouldn't be too hard to update them (hopefully).

RogeredBacon:
Upgrade existing charts to use new module:

Basically the data being provided to the pie charts used to be in separate datasets for each value `[{male, 50}, {female, 72}]`.

Now they require the data be all in one dataset `[{[male, female], [50, 72]}]` which makes more sense.

Fixes #275


astroash:
@RogeredBacon have you rebased master?

RogeredBacon:
I rebased to your fix branch earlier

react-backend/client/src/components/admin/adminUsersPage.js
```diff
@@ -85,23 +83,43 @@ export default class AdminUsersPage extends Component {
   getActivitiesForChart = (activities) => {
     if (!activities) return [];
 
-    return activities.map(({ name, count }, index) => ({
-      value: count,
-      color: this.getColorPair(index).color,
-      highlight: this.getColorPair(index).highlight,
-      label: name,
-    }));
+    const activitiesData = {
```
astroash: I'd recommend trying some other array methods here as you can minimise the amount of code you write. You can do something like this:

```js
    const primaryColors = ['#F7464A', '#46BFBD', '#FDB45C', '#949FB1', '#4D5360'];
    const secondaryColors = ['#FF5A5E', '#5AD3D1', '#FFC870', '#A8B3C5', '#616774'];

    const activitiesData = {
      labels: activities.map(el => el.name),
      datasets: [
        {
          data: activities.map(el => el.count),
          backgroundColor: primaryColors,
          hoverBackgroundColor: secondaryColors,
        },
      ],
    };
```

I've also added in an array of colours needed for `chartjs2` and removed the label in datasets as it's not needed 😊

react-backend/client/src/components/admin/adminUsersPage.js
```diff
@@ -85,23 +83,43 @@ export default class AdminUsersPage extends Component {
...
   getGendersForChart = (genders) => {
     if (!genders) return [];
 
-    return genders.map(({ sex, count }, index) => ({
-      value: count,
-      color: this.getColorPair(index).color,
-      highlight: this.getColorPair(index).highlight,
-      label: sex,
-    }));
+    const genderData = {
```
astroash: See above, also needs colours 🌈

react-backend/client/src/components/admin/adminUsersPage.js
```diff
@@ -146,18 +164,32 @@ export default class AdminUsersPage extends Component {
...
   getAgeGroupsForChart = (ageGroups) => {
     if (!ageGroups) return [];
-    return ageGroups.map(({ agegroups, agecount }, index) => ({
-      value: agecount,
-      color: this.getColorPair(index).color,
-      highlight: this.getColorPair(index).highlight,
-      label: agegroups,
-    }));
+
+    const ageData = {
```
astroash: 🌈 Colour me bruv...

react-backend/client/src/components/admin/adminUsersPage.js
eliascodes: Colors should go in and/or come from the style guide.

react-backend/client/src/components/admin/adminUsersPage.js
```diff
@@ -85,23 +83,43 @@ export default class AdminUsersPage extends Component {
...
+      datasets: [
+        {
+          data: [],
+          label: [],
+        },
+      ],
+    };
+
+    activities.forEach(({ name, count }) => {
+      activitiesData.datasets[0].data.push(count);
```
eliascodes: Is this the correct shape of the data? `activitiesData` is passed in on line 37:

https://github.com/TwinePlatform/DataPower/blob/5bc5e8b413a328d9740a56644f4fedb3f21956e9/react-backend/client/src/components/admin/adminUsersPage.js#L37

The API documentation says that `activitiesNumbers` on this line looks like `[{ name, count},...]`. There's nothing about a dataset field, where's that coming from?

RogeredBacon: - By the Adapter Pattern do you mean pulling out the code to a separate class/file?

- I think I'm going to remove colours for now anyway as its something we'll need to properly think about later.

-The chart api needs all data to be within a datasets field to be able to work properly especially as we're using labels as the legend.

eliascodes: > The chart api needs all data to be within a datasets field to be able to work properly especially as we're using labels as the legend.

Sorry, my mistake, I skipped over the lines where you initialised your datastructure. Nevermind.

> By the Adapter Pattern do you mean pulling out the code to a separate class/file?

I had thought about pulling all the conversion code out into the API client, but have since changed my mind, since it's far to heavily coupled to chart.js, which is specific to a few components.

Basically, ignore my previous comments.

react-backend/client/src/components/admin/adminUsersPage.js
RogeredBacon: I think I'm gonna remove colours entirely for now, as its quite an important thing I feel we need to discuss:
- What colours?
- Should colours for gender and age group graphs be the same every time?
- There can be an infinite amount of activities so should the colours be randomly generated?
- If we do randomly generate colours, should we make sure that each colour is different enough from the others to be easily recognisable?
- Should the colours for gender be the standard blue/red?
- If we don't use the blue/red standard should we actively avoid these colours in the gender graph, to minimise confusion?

eliascodes:
@RogeredBacon would ❤️ it if you rebased 🙏

react-backend/client/src/components/authentication/signupcb.js
```diff
@@ -58,10 +56,16 @@ export default class CBsignup extends Component {
         <h1>Please provide us with required information on your business</h1>
         <form className="Signup" onChange={this.handleChange} onSubmit={this.handleSubmit}>
           <Input question="Business Name" option="org_name" />
-          {errors.orgName && <div className="ErrorText">Business name: {errors.orgName.map(e => <span>{e}</span>)}</div>}
+          {errors.orgName && (
```
astroash: Did anything change in this file or is it just prettier layout?

RogeredBacon: Just prettier, I had like 3 merge conflicts on this file when I rebased, so the formatting is probably different

eliascodes:
@RogeredBacon are you planning on doing the other chart components? They'll be broken now right?

RogeredBacon:
All of the charts work fine now, as far as I can tell. There's only four of them on adminUsersPage + the feedback one Aishas working on.

RogeredBacon:
Thanks for the offer, I'll follow the tutorial and see how it goes first.

RogeredBacon:
Rightio I've done the rebase, the hardest bit was struggling with vim.

react-backend/client/package.json
```diff
@@ -5,18 +5,20 @@
   "dependencies": {
     "axios": "^0.18.0",
     "babel-eslint": "^8.2.2",
-    "chart.js": "^1.1.1",
+    "chart.js": "^2.7.2",
     "normalize.css": "^8.0.0",
     "polished": "^1.9.2",
-    "prop-types": "^15.6.0",
+    "chart.js": "^2.7.2",
```
astroash: You've got a duplicate of `chart.js` here.

RogeredBacon: Oops, fixed.

PR #279, opened 2018-03-14
add styled components + style CB signup page

eliascodes:
Fixes #254 

~~This is a just a small example of how I'm imagining us to organise and style the components once we get the style guide.~~

I've now finished the styling for the CB signup page. The CB signup page contains commented out code that will work once #265 is merged, but this can go in now anyway, since the component isn't hooked up to the rest of the app.

PR #291, opened 2018-03-21
Additional updates to new CbSignupPage

eliascodes:
These were changes that I wanted to include in the previous PR (#279), but it was merged before I could include them.

* The `region` field is set by the change handler anyway, but better to
pre-define it here, as it's more self-documenting.
* Comment API code back in
* Fix up error messages on form validation failure.

### DIFF
--- a/react-backend/client/src/cb-admin/pages/Signup.js
+++ b/react-backend/client/src/cb-admin/pages/Signup.js
@@ -1,7 +1,7 @@
 import React from 'react';
-// import PropTypes from 'prop-types';
-// import { pick, pathOr, equals } from 'ramda';
-// import { CbAdmin } from '../../api';
+import PropTypes from 'prop-types';
+import { pick, pathOr, equals } from 'ramda';
+import { CbAdmin } from '../../api';
 import { Form, FormSection, PrimaryButton } from '../../shared/components/form/base';
 import { Heading, Paragraph, Link } from '../../shared/components/text/base';
 import LabelledInput from '../../shared/components/form/LabelledInput';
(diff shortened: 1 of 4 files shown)


react-backend/client/src/cb-admin/pages/Signup.js
```diff
@@ -68,26 +68,27 @@ export default class CbAdminSignup extends React.Component {
...
-    //       this.history.push('/error/unknown');
-
-    //     }
-    //   });
+
+    CbAdmin.create(payloadFromState(this.state))
+      .then(() => {
+        this.props.history.push('/cb/login?ref=signup');
+      })
+      .catch((error) => {
```
astroash: Wanna use my weird promise hack to avoid accessing `this.history` after the component has unmounted?

eliascodes: It would be helpful to post a link so I don't have to go hunting around for this code:

https://github.com/TwinePlatform/DataPower/blob/077f7e5a18de50ba93fe8a9e6b0c5c1ef3f8ecef/react-backend/client/src/components/hoc/admin_login_check.js

eliascodes: In any case, I think I'm probably gonna cover this in a separate HOC

eliascodes:
> Also, is there a reason for this Component isn't in action?

This isn't a sentence, I don't know what you're asking.

astroash:
☝️ Is there a reason this component isn't being used in by react router?

eliascodes:
Hooked up the component to the router

PR #300, opened 2018-03-26
Form validation hoc

eliascodes:
So I've added a HOC to handle this form validation business because we're doing very similar things in several places.

The basic idea is:
1. Write your `<form>` component.
2. Write your submit handler as a promise-returning function _without_ any error handling or `e.preventDefault()`
3. The promise-returning function (☝️) should accept as an argument an object containing all the input values in the form
4. Write an object where the keys are HTTP status codes, and the values are URLs to redirect to if the form submission returns with that status code.
5. Give all that stuff to the HOC creator.
6. ...??
7. Profit.

For an example, see `/cb-admin/pages/Signup.js` in this PR.

Not necessarily urgent, but nice IMO.


react-backend/client/src/shared/components/form/withValidation.js
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    onChange = (e) => {
+      this.setState({ [e.target.name]: e.target.value });
+    }
+
+    onSubmit = (e) => {
+      e.preventDefault();
+
+      const fdata = new FormData(e.target);
```
astroash: `const` name could be a bit more explicit

react-backend/client/src/cb-admin/pages/Signup.js
```diff
@@ -36,139 +39,101 @@ const regions = [
...
+      </FormSection>
+
+      <FormSection flexOrder={3}>
+        <PrimaryButton>Submit</PrimaryButton>
+      </FormSection>
+
+      <FormSection flexOrder={4}>
+        <Paragraph>
+          Already a subscriber? <Link to="/cb/login">Login</Link>
+        </Paragraph>
```
RogeredBacon: I think this should probably be /logincb

astroash:
@eliascodes & @RogeredBacon do we still want to set some time aside this week to talk through this PR?

eliascodes:
@astroash sure, whenever

astroash:
I can either do after the call with Sonja or Thursday.

astroash:
@eliascodes shall we get this merged in?

eliascodes:
Yeah, let me just manually test it again, I may have fucked the rebase a little.

eliascodes:
Will mark it "awaiting review" when I'm sure it still works.

eliascodes:
Cool, it still works, have at it.

PR #310, opened 2018-03-28
Cb admin dashboard

eliascodes:
CB admin dashboard styled + dots added to all pages.

Buttons on the dashboard are slightly off center, but at this stage, CBA, would rather finish the other views.

I decided against changing the auth system; I started, but it turned into this huge thing, so better leave it till later.


react-backend/client/src/shared/components/Dots.js
```diff
@@ -0,0 +1,31 @@
+import styled from 'styled-components';
+
+
+export default styled.div`
+  z-index: -1;
+  left: 50%;
+  top: 0;
+  transform: translate3d(-50%, 0, 0);
+
+  &, &::before, &::after {
```
RogeredBacon: What does this do?

eliascodes: I don't know, I just copied it from their website 🤷‍♂️

RogeredBacon: Haha coding in a nutshell

react-backend/client/src/App.js
```diff
@@ -56,154 +57,156 @@ class App extends Component {
...
-            path="/admin/feedback"
-            updateLoggedIn={this.updateLoggedIn}
-            component={adminLoginCheck(CbAdminFeedbackPage)}
-          />
-
-          <Route exact path="/internalServerError" component={InternalServerError} />
-          <Route component={NotFound} />
-        </Switch>
-      </Container>
+      <Dots>
```
astroash: How come there's such a big dif here? Can this be avoided as this caused me some major rebase woes last time.

eliascodes: An extra wrapping element means the indentation level for all children increases by one.

eliascodes: What changes did you make to the App.js? Just swap out some components in the route definitions? If so, handling conflicts on rebase is simple: accept the current changes (i.e. changes _not_ coming from your commits), then past your components back in.

PR #319, opened 2018-03-30
Cb profile

eliascodes:
Adds the CB profile settings page, check commits for deets.

Diff looks bigger than it really is


react-backend/client/src/App.js
```diff
@@ -33,6 +31,8 @@ import ConfirmPassword from './cb-admin/pages/ConfirmPassword';
 import ForgotPassword from './cb-admin/pages/ForgotPassword';
 import ResetPassword from './cb-admin/pages/ResetPassword';
 import CbDashboard from './cb-admin/pages/Dashboard';
+import CbAdminFeedbackPage from './cb-admin/pages/FeedbackPage';
+import CbSettingsPage from './cb-admin/pages/Settings';
```
astroash: Should this be called `CbAdminSettingsPage`? I thought this was the naming convention we were going with..

react-backend/client/src/App.js
```diff
@@ -33,6 +31,8 @@ import ConfirmPassword from './cb-admin/pages/ConfirmPassword';
 import ForgotPassword from './cb-admin/pages/ForgotPassword';
 import ResetPassword from './cb-admin/pages/ResetPassword';
 import CbDashboard from './cb-admin/pages/Dashboard';
+import CbAdminFeedbackPage from './cb-admin/pages/FeedbackPage';
```
astroash: Does this work? You've renamed this component to `FeedbackPage` lower down.

react-backend/client/src/shared/components/layout/base.js
```diff
@@ -13,4 +13,5 @@ export const FlexContainerCol = styled(FlexContainer)`
 export const FlexContainerRow = styled(FlexContainer) `
   flex-direction: row;
   flex-wrap: wrap;
+  ${props => (props.flex ? css`flex: ${props.flex};` : '')}
```
astroash: What does this do?

react-backend/shared/models/query_builder.js
```diff
@@ -0,0 +1,93 @@
+/*
+ * Query builder utilities
```
astroash: Is this needed? I feel like it is taking away from the readability of our code.

eliascodes: Don't you think it's nice to be able to generate simple queries dynamically? I can imagine we might want to have several slightly different views on the same data, which would currently require us to write one function for each view, instead of just making several calls to the same function:
```js
select('cbusiness', ['name', 'yob', 'gender'], { id: 1 });
select('cbusiness', ['email'], { is_email_contact_consent_granted: true });
select('cbusiness', ['phone_number'], { is_sms_contact_consent_granted: true });
// ... etc
```
You don't think the above is readable?

react-backend/client/src/shared/components/layout/base.js
eliascodes: It sets the `flex` property only if the `flex` prop is set

react-backend/client/src/App.js
eliascodes: Where?

react-backend/client/src/App.js
eliascodes: Yeah, sure, I'll rename them all in a future PR.

react-backend/client/src/App.js
astroash: Here:
https://github.com/TwinePlatform/DataPower/blob/3edf17ab86f1902c4740a249db1c0d533323a7f4/react-backend/client/src/cb-admin/pages/FeedbackPage.js#L85

It's being weird when I pull this branch and not registering the git mv/rename

eliascodes: `import FOO from 'bar'` is not the same as `import { FOO } from 'bar'`. This works fine as far as I can see.

astroash: Okay sweet, in that case ignore me.

react-backend/shared/models/query_builder.js
astroash: I'm not completely convinced but happy to go with it 😊

eliascodes:

PR #317, opened 2018-03-30
💄 Styled visitor homepage

astroash:
- also added Heading2 component to text/base.js

Fixes #313 

Looks like:
screen shot 2018-03-30 at 8 45 35 pm

PR #321, opened 2018-03-31
💄 Style Visitor activities

astroash:
**#317 to be merged first! Base set to visitor-homepage to see changes**

Looks like:
screen shot 2018-03-31 at 11 28 48 am


react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -2,9 +2,39 @@
...
+`;
+
+const SnakeContainerRow = FlexContainerRow.extend`
+  width: 100%;
+  justify-content: space-between;
+  &:nth-child(2n) {
+  flex-direction: row-reverse;
+  `;
+
+const capitaliseFirstName = name => name.split(' ')[0].replace(/\b\w/g, l => l.toUpperCase());
```
eliascodes: First name is first, so this is just capitalising the first character of a string:
```js
const capitalise = str => str.slice(0, 1).toUpperCase().concat(str.slice(1));
```

react-backend/client/src/components/visitors/qrprivacy.js
```diff
@@ -1,30 +1,27 @@
...
-    <img className="Privacy__image" src="/images/question.png" alt="question" />
-    <h3 className="Privacy__title">What information are we storing?</h3>
-    <p className="Privacy__text">
+const ATag = styled.a`
+`;
+const qrPrivacy = () => (
+  <aside>
+    <h3>What information are we storing?</h3>
+    <Paragraph>
       Your name, sex, age, and reason for visiting today, as well as the time and date of your
```
eliascodes: "`Your name, **gender**, age and ...`"

react-backend/client/src/components/visitors/qrprivacy.js
```diff
@@ -1,30 +1,27 @@
...
       visit.
       <br />
       <br />
-    </p>
-    <h3 className="Privacy__title">Why are we collecting this information?</h3>
-    <p className="Privacy__text">
+    </Paragraph>
+    <h3>Why are we collecting this information?</h3>
```
eliascodes: indentation

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -131,23 +161,36 @@ export default class QRCode extends Component {
...
-              onClick={this.changeActivity}
-            />
-          ))}
-        </section>
-        <QRPrivacy className="col-3" />
-      </div>
+      <section>
+        <StyledNav>
+          <Heading>
+              Welcome back, {capitaliseFirstName(this.state.username)}! Why are you here today?
```
eliascodes: indentation

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -2,9 +2,39 @@
...
+  justify-content: center;
+`;
+
+const HalfFlexContainerRow = FlexContainerRow.extend`
+  width: 50%;
+  padding: 1rem;
+  flex-wrap: wrap;
+`;
+
+const SnakeContainerRow = FlexContainerRow.extend`
```
eliascodes: Could you quickly explain why this is necessary?

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -131,23 +161,36 @@ export default class QRCode extends Component {
...
-      </div>
+      <section>
+        <StyledNav>
+          <Heading>
+              Welcome back, {capitaliseFirstName(this.state.username)}! Why are you here today?
+          </Heading>
+        </StyledNav>
+        <StyledSection>
+          <HalfFlexContainerRow>
+            {this.state.activities.map((activity, index) => (
```
eliascodes: I think the UI semantic we're actually trying to express here is "row of buttons with alternating styles".

I think a better way to structure that would be to have an `AlternatingComponentRow`, which would accept as props the two components to render in alternating fashion, as well as a data array which would be passed as props to those two components.

```js
<AlternatingComponentRow 
  one={PrimaryButton}
  two={SecondaryButton}
  data={[ { content: 'Violin', onClick: () => {}, ... }, { content: 'Cafe', ... }, ... ]}
/>
```
Then then all the messy logic can be internal to that.

react-backend/client/src/components/visitors/qrprivacy.js
```diff
@@ -1,30 +1,27 @@
...
+    <Paragraph>
       We are always trying to improve what we offer you. By signing in, you are helping us make
       our services better.
-    </p>
+    </Paragraph>
+    <ATag href="http://www.powertochange.org.uk/data-protection-funding-applications/">
+      Data Protection Policy
```
eliascodes: There's already `Link` styled tag in the text base, this is not in keeping with the designs.

react-backend/client/src/components/visitors/qrprivacy.js
```diff
@@ -1,30 +1,27 @@
...
       visit.
       <br />
       <br />
-    </p>
-    <h3 className="Privacy__title">Why are we collecting this information?</h3>
-    <p className="Privacy__text">
+    </Paragraph>
+    <h3>Why are we collecting this information?</h3>
+    <Paragraph>
       We are always trying to improve what we offer you. By signing in, you are helping us make
```
eliascodes: All the text in this component is black, which is not in keeping with the design. Needs to be grey. Refer to the style guide component.

react-backend/client/src/components/visitors/qrcode.js
astroash: Names are given as a full string eg `"Jessica Jones"`, so I also need to drop the second name. I went for regex over slice as it was marginally less code over `.slice`.

react-backend/client/src/components/visitors/qrcode.js
astroash: This is creating a div  which grabs the second child and reverses it's direction. When the divs are stacked on top of each other the elements order is concertinaed like a snake.

react-backend/client/src/components/visitors/qrcode.js
astroash: The way I have done this the buttons are always created in the same order, but use flex to switch their ordering when displayed, rather than creating each 2nd row in reveres. The benefit of this is on smaller screen sizes the buttons will appear in the correct order if stacked in 1 column.

Agree that there's probably a better way of writing this. HMU next time you're working I'd find it useful to chat through it.

react-backend/client/src/components/visitors/qrprivacy.js
astroash: This is an external link. I assumed you couldn't use `Link` from `react-router-dom` for external links.

eliascodes: Ah, true. Fair enough, but according to the design, it still needs to look like the other links.

react-backend/client/src/components/visitors/qrcode.js
eliascodes: Yeah ok I don't want this changed at the moment, but we can talk about it in future.

react-backend/client/src/components/visitors/qrprivacy.js
astroash: I'll nab your styling from `Link`

react-backend/client/src/components/visitors/qrprivacy.js
astroash: I didn't notice all the text is meant to be grey. Can I add this to everthing in `/shared/components/text/base.js`?

eliascodes: I guess so

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -16,7 +16,7 @@ const StyledNav = styled.nav`
 `;
 
 const StyledSection = styled.section`
-  margin: 3rem 0;
+  margin: {props => props.margin 'rem 0'}
```
eliascodes: Isn't this a syntax error?

astroash: Na, it sticks the number passed in from props in front of the string.

eliascodes: Nope, definitely a syntax error.

screen shot 2018-04-02 at 14 19 14

It's not crashing because you're not doing string interpolation properly. You're missing the `$` in front of the `{`.

astroash: Oops

astroash:
☝️ I've made the two columns responsive, so they change width based on screensize, as 60/40 was looking weird on desktop. It might be worth us also making the default main body margin responsive if we're going to start fine tuning layout for different devices.

eliascodes:
pls resolve conflicts

PR #322, opened 2018-03-31
Visitor details page

eliascodes:
**MERGE #319 FIRST**

Visitor details page, including printable view. Deets in commits.

### DIFF
--- a/react-backend/client/src/cb-admin/pages/Visitor.js
+++ b/react-backend/client/src/cb-admin/pages/Visitor.js
@@ -1,0 +1,301 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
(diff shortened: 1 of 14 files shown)


react-backend/client/src/cb-admin/pages/Visitor.js
```diff
@@ -0,0 +1,301 @@
...
+          </Form>
+        </FlexContainerRow>
+      </Col>
+    );
+  }
+
+  render() {
+    return (
+      <React.Fragment>
+        {
```
astroash: Is this rendering both versions of the page, but PrinterFriendly is set to display none?

eliascodes: Yes

eliascodes:

PR #323, opened 2018-04-01
Visitor details

eliascodes:
**MERGE #322 FIRST**

Visitor details page, not much changed.

screen shot 2018-04-01 at 17 03 43

screen shot 2018-04-01 at 17 04 07

### DIFF
--- a/react-backend/client/src/cb-admin/pages/VisitorDetails.js
+++ b/react-backend/client/src/cb-admin/pages/VisitorDetails.js
@@ -1,0 +1,225 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import moment from 'moment';
+import { assocPath, filter, invertObj, map, project, sortBy } from 'ramda';
+import { FlexContainerCol, FlexContainerRow } from '../../shared/components/layout/base';
+import { Heading, Link } from '../../shared/components/text/base';
+import { Form as Fm } from '../../shared/components/form/base';
+import LabelledSelect from '../../shared/components/form/LabelledSelect';
+import TranslucentTable from '../components/TranslucentTable';
+import { Visitors } from '../../api';
+
+
+const Nav = styled.nav`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  flex: 1;
+`;
+
+const FlexItem = styled.div`
+  flex: ${props => props.flex || '1'};
+  height: 100%;
+`;
+
+const Form = Fm.extend`
+  width: 100%;
+`;
+
+const Row = FlexContainerRow.extend`
+  align-content: center;
+  align-items: flex-start;
+  flex: 3;
+`;
+
+const HyperLink = Link.extend`
+  flex: ${props => props.flex || '1'};
+`;
+
+const FormSection = styled.section`
+  width: 20%;
+`;
+
+
+const keyMap = {
+  visit_id: null,
+  visitor_id: 'Visitor ID',
+  gender: 'Gender',
+  yob: 'Year of birth',
+  activity: 'Activities',
+  visit_date: 'Date of visit',
+};
+
+const colToState = invertObj(keyMap);
+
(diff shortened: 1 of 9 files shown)


react-backend/client/src/cb-admin/pages/VisitorDetails.js
```diff
@@ -0,0 +1,225 @@
...
+  yob: 'Year of birth',
+  activity: 'Activities',
+  visit_date: 'Date of visit',
+};
+
+const colToState = invertObj(keyMap);
+
+const columns = Object.values(keyMap).filter(Boolean);
+
+const sortOptions = [{ key: '0', value: '' }].concat(columns.map((col, i) => ({ key: `${i + 1}`, value: col })));
```
astroash: What's going on here?

react-backend/client/src/cb-admin/pages/VisitorDetails.js
```diff
@@ -0,0 +1,225 @@
...
+        this.props.updateAdminToken(res.headers.authorization);
+        const visits = res.data.result;
+
+        const order = visits.map(v => v.visit_id);
+        const items = visits.reduce((acc, v) => {
+          acc[v.visit_id] = { ...v, display: true };
+          return acc;
+        }, {});
+
+        this.setState(assocPath(['visits'], { order, items }));
```
astroash: What's the benefit of using `assocPath`?

react-backend/client/src/cb-admin/pages/VisitorDetails.js
```diff
@@ -0,0 +1,225 @@
...
+      const displayForGender = genderFilter ? item.gender === genderFilter.toLowerCase() : true;
+      let displayForAge = true;
+
+      if (ageFilter) {
+        const age = currentYear() - item.yob;
+        const match = ageFilter.match(/(\d{1,2})[-+](\d{0,2})/);
+
+        if (match) {
+          const [_, min, max] = match; // eslint-disable-line
+          displayForAge = age > +min && age < (+max || Infinity);
```
astroash: What's this doing?

react-backend/client/src/cb-admin/pages/VisitorDetails.js
eliascodes: `keyMap` describes the mapping between keys in state and column display names (hence `visit_id` being `null`).

`colToState` is the reverse mapping.

`columns` is a list of the column display names.

`sortOptions` is the list of things we can sort the table by (i.e. the columns). But in order to be usable by `LabelledSelect` they need to be in a particular form. Also we want a default empty option at the top.

react-backend/client/src/cb-admin/pages/VisitorDetails.js
eliascodes: Using the functional form of `setState` is recommended because state updates are asynchronous.

Also (the real reason): I temporarily forgot how to do this with object spreading:
```js
this.setState(state => ({ ...state, visits: { ...state.visits, order, items } }));
```
Interested to know which you think is more readable (assuming the reader understands what `assocPath` does).

react-backend/client/src/cb-admin/pages/VisitorDetails.js
eliascodes: Setting a flag that determines whether the item in question has `yob` field that is compatible with the age filter.

Age filter is specified as `X-Y` or `Z+`, so I extract the numbers via regex, then do the comparison to set the flag.

Frankly it's a little slapdash, there's definitely higher order concepts in there, but it'll do for now.

eliascodes:

eliascodes:
@RogeredBacon sure, no problem. If you can have a think about which bits exactly you'd like more explanation about, that'd be really helpful.

eliascodes:
@RogeredBacon could you merge please? @astroash has already gone through this PR, and I don't want to delay this until Thursday. Thanks.

RogeredBacon:
Thanks, will do

PR #326, opened 2018-04-02
🐛 Fix instacam problems

### LINKED ISSUE #325: 🐛 Intermittent qr code error
Error is sometimes displayed after scanning QR code....

screen shot 2018-04-02 at 1 39 22 pm

astroash:
I'm not sure if this is the whole problem solved, but I can't replicate the error...

- Add ref to div rather than grabbing from DOM
- Comment out iife as I _think_ its duplicate code

Fixes #325

### DIFF
--- a/react-backend/client/src/components/visitors/qrcode.js
+++ b/react-backend/client/src/components/visitors/qrcode.js
@@ -131,7 +131,6 @@ export default class QRCode extends Component {
             reject('failure stopping scanner');
           });
       } else {
-        this.previewDiv.play();
         reject('ADD LISTENER FAILED');
       }
     });
@@ -144,9 +143,7 @@ export default class QRCode extends Component {
           reject('ERROR HAPPENING AT getCameras');
         }
       })
-      .catch((err) => {
-        reject(err);
-      });
+      .catch(reject);
   })
 
   handleVideo = () => this.setState({ login: this.state.login + 1 });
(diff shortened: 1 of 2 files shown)


react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -93,6 +63,41 @@ export default class QRCode extends Component {
...
+        scanner
+          .stop()
+          .then(() => {
+            resolve(content);
+          })
+          .catch(() => {
+            reject('failure stopping scanner');
+          });
+      } else {
+        this.previewDiv.play();
```
eliascodes: If `previewDiv` is `null` we still try to call a method on it? This is what caused the original bug.

react-backend/client/src/components/visitors/qrcode.js
```diff
@@ -93,6 +63,41 @@ export default class QRCode extends Component {
...
+
+    Instascan.Camera.getCameras()
+      .then((cameras) => {
+        if (cameras.length > 0) {
+          scanner.start(cameras[0]);
+        } else {
+          reject('ERROR HAPPENING AT getCameras');
+        }
+      })
+      .catch((err) => {
```
eliascodes: `catch(reject)`

eliascodes:

eliascodes:
@astroash fix conflicts pls

PR #330, opened 2018-04-03
CB Activities page

eliascodes:
Adds CB activities page. Code is gross, but more important to get this in right now.

Also deleted a bunch of superseded code, some of which should have been removed in some of my previous PRs.


RogeredBacon:
Yeah it looks like the checkboxes definitely don't work for me for some reason, even though the delete buttons are fine. I tried Chrome as well as Firefox but no luck.

I raised an issue as you suggested #331

react-backend/client/src/cb-admin/pages/Activities.js
```diff
@@ -0,0 +1,267 @@
...
+    Activities.get(this.props.auth)
+      .then((res) => {
+        const activities = res.data.result;
+
+        const order = activities.map(activity => activity.id);
+        const items = activities.reduce((acc, activity) => {
+          acc[activity.id] = activity;
+          return acc;
+        }, {});
+        this.setState(state => ({ ...state, activities: { items, order } }));
```
astroash: Are you spreading state here because activities is a nested obj?

react-backend/client/src/cb-admin/pages/Activities.js
```diff
@@ -0,0 +1,267 @@
...
+            return this.props.history.push('/internalServerError');
+
+          default:
+            return this.props.history.push('/admin/login');
+        }
+      });
+  }
+
+  onChange = e =>
+    this.setState(assocPath(['form', e.target.name], e.target.value))
```
astroash: In future, I reckon it's more readable not using `assocPath` in `setState`

eliascodes: Depends how deeply the state is nested. Object spreading can get very unreadable.

react-backend/client/src/cb-admin/pages/Activities.js
eliascodes: Yes.

PR #311, opened 2018-03-28
Visitors Signup Flow - Implement Redesign

RogeredBacon:
> Ok so the formatting is pretty much done except for two things:
> 
> 1. The years function in main.js returns an array like: `[2018, 2017, 2016, ...]`.
> However the LabelledSelect Component needs an array like: `[{key: 0, value: 2018}, {key: 1, value: 2017}, {key: 2, value: 2016}, {..., ...}, ...]`
> 
>     I'm pretty bad at manipulating objects so my question is would it be best to keep the original     array and modify it in some way (via reduce or something); or just create the object from scratch?
> 
> 2. I'm not sure what to do with the checkboxes on the signup page, they're meant to look like this:
> visitor signup page
> 
> Currently they look like this (I know they should be on the right): 
> screenshot-2018-4-4 data power
> 
> The issue I'm having is that the designs want the checkboxes to be inside of the `<p>`, which is fine but any change of the screen size means they look crap as they move around inside the text or randomly out of the bottom. I can keep messing with it till it looks ok most of the time, but I was wondering whether there is a better way of dealing with this?

Should be all good except for the organisation name bit which will follow in another PR. 

Relates #302 
Fixes #303 #312

PR #335, opened 2018-04-05
Add OrgName to Visitor Signup

### LINKED ISSUE #274: Automatically get the name of the CB to display on signup page.
In the signup_form.js file there's a string of XXXXXXXXX which currently represents where we will want the CB's name to go.

This will probably include creating a new query and route, or the modification of an existing one so that we have the CB name available.

RogeredBacon:
This makes the organisations name appear when a visitor is signing up.

Adds small route that just grabs from the middleware minor changes to main.js and signup_form.js

**Must be merged after #311.**

Fixes #274

### DIFF
--- a/react-backend/client/src/api/index.js
+++ b/react-backend/client/src/api/index.js
@@ -171,11 +171,6 @@ export const Visitors = {
       { headers: { Authorization: tkn } },
     );
   },
-
-  getCbName: tkn => axios.get(
-    '/api/users/cb-name',
-    { headers: { Authorization: tkn } },
-  ),
 };
 
 
@@ -186,6 +181,11 @@ export const CbAdmin = {
       {},
       { headers: { Authorization: tkn } },
     ),
+
+  getCbName: tkn => axios.get(
+    '/api/users/cb-name',
+    { headers: { Authorization: tkn } },
+  ),
 
   create: ({ orgName, category, email, password, passwordConfirm }) =>
     axios.post(
--- a/react-backend/client/src/api/index.js
+++ b/react-backend/client/src/api/index.js
@@ -182,7 +182,7 @@ export const CbAdmin = {
       { headers: { Authorization: tkn } },
     ),
 
-  getCbName: tkn => axios.get(
+  __DEPRECATED_get: tkn => axios.get( // eslint-disable-line
     '/api/users/cb-name',
     { headers: { Authorization: tkn } },
   ),
--- a/react-backend/client/src/api/index.js
+++ b/react-backend/client/src/api/index.js
@@ -172,12 +172,10 @@ export const Visitors = {
     );
   },
 
-  getCbName: tkn =>
-    axios.get(
-      '/api/users/cb-name',
-      { headers: { Authorization: tkn } },
-    ),
-
+  getCbName: tkn => axios.get(
+    '/api/users/cb-name',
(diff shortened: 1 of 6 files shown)


react-backend/routes/cb_name_check.js
```diff
@@ -0,0 +1,10 @@
+const router = require('express').Router();
+const cbDetails = require('../database/queries/cb/cb_details');
+
+router.get('/', (req, res, next) => {
+  cbDetails(req.app.get('client:psql'), req.auth.cb_id)
```
astroash: Here you don't need to do a db query as both of the pieces of info you're after are added onto `req.auth` using middleware:

https://github.com/TwinePlatform/DataPower/blob/9ac8d1934aced9d2412c9b9943a4ea89d0f613ab/react-backend/routes/cbauthentication/mw_is_authenticated.js#L14-L23

react-backend/client/src/api/index.js
```diff
@@ -172,6 +172,10 @@ export const Visitors = {
     );
   },
 
+  getCbName: tkn => axios.get(
```
eliascodes: Also this is in the wrong place. Methods in these objects are not organised by _who_ uses them, but by _what_ they provide information about. This is providing information _about_ the CB, and therefore should go in `CbAdmin`, under `get` since it's performing a read operation.

If you look in the existing `CbAdmin.get` method, you'll see a call to `/api/cb/details`, which provides an object describing the CB, including the name, which is what you're looking for. Why not just use that method instead of creating a whole new route?

When developing HTTP APIs like ours, we should seek to design them such that they can comprehensively describe the entities that we are modelling in a concise, but flexible set of endpoints. We should not seek to add an additional route for every single use case if it can be avoided. That is a recipe for a vast and un-maintainable API.

RogeredBacon: I looked at that but I needed an admin token to be able to use that route; I could have added the same route again in the router but checking mwIsAuthenticated instead but I thought that might have been bad practice? I'll move the method to be under CbAdmin.

react-backend/router.js
```diff
@@ -47,6 +48,7 @@ router.use('/visit/check', mwIsAuthenticated, visitorCheck);
 router.use('/visit/add', mwIsAuthenticated, visitInsert);
 router.use('/activities/today', mwIsAuthenticated, activitiesToday);
 router.use('/admin/login', mwIsAuthenticated, adminLogin);
+router.use('/users/cb-name', mwIsAuthenticated, cbName);
```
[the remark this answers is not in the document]
RogeredBacon: Ah I had meant to but forgotten, I'll change it.

react-backend/client/src/components/visitors/main.js
```diff
@@ -27,14 +108,34 @@ export default class Main extends Component {
...
+
+  onClickPrint = () => {
+    window.print();
+  };
+
   setError(messagesArray) {
     this.setState({ error: messagesArray });
   }
 
+  headers = new Headers({
```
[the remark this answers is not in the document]
RogeredBacon: I'll remove it

react-backend/routes/cb_name_check.js
RogeredBacon: Sweet, will do

PR #342, opened 2018-04-06
🐛 fix broken cb signup form

eliascodes:
Fix broken signup form

### DIFF
--- a/react-backend/client/src/api/index.js
+++ b/react-backend/client/src/api/index.js
@@ -2,7 +2,7 @@
  *
  */
 import axios from 'axios';
-import { head, pathOr, equals } from 'ramda';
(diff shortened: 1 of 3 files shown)


RogeredBacon:
screenshot-2018-4-6 data power

Are the error messages meant to look like this? If I was a user I'd have no idea what they were talking about.

eliascodes:
@RogeredBacon fair point, will provide better ones.

eliascodes:
@RogeredBacon try now

RogeredBacon:
Nice, the error messages are great! However I'm getting a 409 error now, when I try to register a CB with the same email as one already registered, and then redirected to the error/unknown 404 page.

eliascodes:
Fixed

PR #344, opened 2018-04-06
🐛 Fix adding duplicate activities

### LINKED ISSUE #272: Can insert duplicate activities
There's no check on whether the activity we're trying to add already exists for that community business.

Related: the query run for the data on the `/admin/users` page doesn't filter out "deleted" activities, is this deliberate @RogeredBacon?

astroash:
- Add query to check if it exists
- with tests
- Display error message on frontend
- Forward if any requests are made when admin token is expired
- Add error styled component to text base
Fixes #272

### DIFF
--- a/react-backend/client/src/cb-admin/pages/Activities.js
+++ b/react-backend/client/src/cb-admin/pages/Activities.js
@@ -3,12 +3,12 @@ import styled from 'styled-components';
 import PropTypes from 'prop-types';
 import { assocPath, dissoc, invertObj } from 'ramda';
 import { FlexContainerCol } from '../../shared/components/layout/base';
-import { Paragraph, Heading, Link as HyperLink } from '../../shared/components/text/base';
+import { Paragraph, Heading, Link as HyperLink, ErrorParagraph } from '../../shared/components/text/base';
 import { Form, PrimaryButton } from '../../shared/components/form/base';
 import LabelledInput from '../../shared/components/form/LabelledInput';
 import LabelledSelect from '../../shared/components/form/LabelledSelect';
 import Checkbox from '../components/Checkbox';
-import { Activities } from '../../api';
+import { Activities, ErrorUtils } from '../../api';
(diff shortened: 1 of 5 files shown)


react-backend/client/src/cb-admin/pages/Activities.js
```diff
@@ -170,23 +181,29 @@ export default class ActivitiesPage extends React.Component {
...
 
   render() {
     const { errors } = this.state;
-
+    const errorMessage = <ActivitiesError> {errors.general} </ActivitiesError>;
     return (
       <FlexContainerCol expand>
         <Nav>
-          <HyperLink to="/admin"> Back to dashboard </HyperLink>
+          <FlexItem>
```
eliascodes: Why does this need to be wrapped?

react-backend/routes/activities_add.js
```diff
@@ -1,15 +1,30 @@
...
 
-router.post('/', (req, res, next) => {
+router.post('/', validate(schema),async (req, res, next) => {
   const activityToAdd = req.body.name;
   const pgClient = req.app.get('client:psql');
+  try {
+    const exists = await activityCheckExists(pgClient, activityToAdd, req.auth.cb_id);
+    
+    if (exists) {
+      return res.status(409).send({ error: 'Activity already exists' });
```
eliascodes: Use `Boom` instead

react-backend/routes/activities_add.js
```diff
@@ -1,15 +1,30 @@
...
 
-const router = express.Router();
+const schema = {
+  body: {
+    name: Joi.string().required(),
+  },
+};
 
-router.post('/', (req, res, next) => {
+router.post('/', validate(schema),async (req, res, next) => {
```
eliascodes: Space?

react-backend/client/src/cb-admin/pages/Activities.js
astroash: Otherwise the title wasn't centered.

PR #347, opened 2018-04-07
🚀 read port from environment

eliascodes:
Heroku requires apps to bind to the port set in the `$PORT` environment variable, but we only need to do this for the production environment.


eliascodes:
@astroash you can merge if you like, no point waiting till Monday.

Issue #357, opened 2018-04-09
Proposed new data model


eliascodes:
After going back to #210 and thinking more about the schema, I propose we move to something like the following schema before we launch in earnest to avoid some complicated migrations:

<img width="1148" alt="screen shot 2018-04-09 at 23 39 41" src="https://user-images.githubusercontent.com/12719553/38526740-628d54aa-3c4f-11e8-8f7b-c429fa34bd6a.png">

IMO this better models the actual entities we are dealing with, and provides a more normalised data model.

We have had confusion around naming the "cbusiness" table, because it was actually conflating two separate concepts: the community business itself, and the community business admin. In fact it was conflating more than two concepts by including stuff about the password reset tokens and "genre" (i.e. business sector) in the same table. These are now all split out into their own tables

eliascodes:
It comes from a database normalisation rule called 2nd normal form, which states that all columns in a table must depend ONLY on the primary key (or to put it another way, each column should be some inherent quality of the entity being described).

A business exists apart from it's geographical location (businesses can move, can exist in multiple locations), and also from it's sector (businesses can pivot, expand into other sectors, etc).

The same is true for `activity_category`, but also: it's one-to-many now, but that may not be true in future. Who's to say an activity may not belong to multiple categories?

I'm not really satisfied with the way the activity schedule is captured here. Again, a separate table may be better, but I'm happy making this change in the future via a migration once we better understand the way these things are actually used by the CBs.

eliascodes:
Are there any more opinions or concerns about this?

eliascodes:
Something to note: in the above model, there is no direct relationship between a visitor and a community business. This would allow in principle a single visitor account to register once and then use their QR code at multiple businesses.

However, Sonja mentioned alternative login mechanisms for visitors, including choosing from a list of registered visitors. So how would visitors who have registered but not actually selected an activity (and therefore there is no "visit" record created) be associated with one or more businesses?

astroash:
Sounds like it would be worth having a linking table for `cb_id`'s and `visitor_id`'s. I can totally imagine the scenario where little old Gladis signs up for a visitor account at her local cb and looses her QR code. Without a link between the two cb's won't be able to find her account to resend it to her.

RogeredBacon:
Does this mean that we're allowing any CB to see a list of all registered users, across all CBs on the Visitor Details page?

eliascodes:
[...]

Issue #263, opened 2018-03-08
Architecture theory in docs


astroash:
@eliascodes with all the sweet architectural magic thats happening at the minute, I was wondering if ~~we~~ you can add an md in the docs with references to some of the theory behind it. I would find it interesting to read up on, kthxbye 😊

PR #364, opened 2018-04-10
Detailed documentation of new API design and Data model

eliascodes:
Related to #357 
Related to #263 

This adds detailed documentation of the proposed new API (with minor changes from the previous version). I've added a couple references for those interested in the design of REST-ish APIs.

Also adds image of new data-model, though this may still change slightly in the near future.


docs/current_api.md
```diff
@@ -60,7 +57,12 @@ This design is based on moving towards an API that is more RESTful, and attempti
 ### Error Response
 All non-successful responses should return appropriate status codes in the range `4xx` or `5xx`, with `"result": null`, and a descriptive message in `"error": { "message": "..." }`. If the error is due to validation, the `"error"` object should have a key of `"validation"` which is an array of strings describing the validation errors.
 
-### Routes
+### Accept and Content-Type Headers
+Clients should set the `Accept` header in order to indicate what content-type(s) they expect as a response. If no `Accept` header is set, the API will default to a `application/json` content-type. The API will set the `Content-Type` header accurately to reflect the response format. The two response formats that should be supported are:
+* JSON
+* XLS (via XML)
```
astroash: Can this be changed to CSV

RogeredBacon: Does this relate to the spreadsheets or http requests?

eliascodes:
@astroash 

> Also, are these all the api doc changes you imaged? When you described it I assumed it was going to be bigger.

I suspect you missed the new file I added because the diff was suppressed. Check again.

docs/new_api.md
```diff
@@ -0,0 +1,646 @@
...
+For example, in `GET /api/visitors`, the response body is:
+```
+{ result: [{ id, name, gender, email, ... }] }
+```
+We should specify a filter on `gender` and `email` in the following way:
+```
+GET /visitors?gender=foo&email=bar
+```
+
+In addition, the following query parameters should be standard:
```
astroash: When you stay they're standard does this mean that every get request should have all of these params?

docs/new_api.md
```diff
@@ -0,0 +1,646 @@
...
+* `fields`: Comma separated list of fields in the response body to return. For example, `GET /visitors?fields=name,gender` returns only the `name` and `gender` field in the response body, instead of the entire object.
+* `sort`: Single field by which to sort results
+* `order`: Order by which to sort field specified by `sort`. One of `asc` or `desc`
+* `since`: ISO86001 date string (with timezone?) specifying the **earliest** time from which results should be returned. This is not always meaningful for every endpoint, and is therefore ignored in those cases.
+* `until`: ISO86001 date string (with timezone?) specifying the **latest** time from which results should be returned. This is not always meaningful for every endpoint, and is therefore ignored in those cases.
```
astroash: ISO 8601 includes a timestamp with a  'z' offset of UCT.

docs/new_api.md
```diff
@@ -0,0 +1,646 @@
...
+}
+```
+
+## Error Response
+All non-successful responses should return appropriate status codes in the range `4xx` or `5xx`, with `"result": null`, and a descriptive message in `"error": { "message": "..." }`. If the error is due to validation, the `"error"` object should have a key of `"validation"` which is an array of strings describing the validation errors.
+
+## Accept and Content-Type Headers
+Clients should set the `Accept` header in order to indicate what content-type(s) they expect as a response. If no `Accept` header is set, the API will default to a `application/json` content-type. The API will set the `Content-Type` header accurately to reflect the response format. The two response formats that should be supported are:
+* JSON
+* XLS (via XML)
```
astroash: Change to CSV

docs/new_api.md
```diff
@@ -0,0 +1,646 @@
...
+
+* `NULL`: Un-authenticated user
+* `CB`: Community business user
+* `CB_ADMIN`: Community business user with elevated admin permissions
+* `TWINE_ADMIN`: Twine platform admin user
+
+The `CB_ADMIN` permission level is a short-lived authorization with broader rights than the `CB` permission level. It is attained via a permission escalation mechanism (currently confirming password).
+
+## Activities
+### `GET    /api/activities/`
```
astroash: I'm not clear on this layout. Is this meant to be a request to this endpoint without params by a CB?

docs/new_api.md
```diff
@@ -0,0 +1,646 @@
...
+    ...
+  },
+  gender: {
+    <GENDER_NAME>: [{ id: String, activity: String, visit_date: String }],
+    ...
+  },
+}
+```
+
+## Design Resources
```
astroash: 😍

docs/current_api.md
astroash: ☝️

docs/new_api.md
eliascodes: It means that you should be able to pass these parameters to every endpoint and for them to have the desired effect if the parameter is meaningful for that endpoint.

docs/new_api.md
eliascodes: It's meant to be a request to that endpoint by a client that has a token corresponding to a `CB` role.

docs/new_api.md
astroash: Surely it's not necessary to write functionality for this is on all `GET` requests. For example on the feedback request will never need to be sorted as there is no tabulated data.

eliascodes: We write a function to do this once, and can use it everywhere.

docs/new_api.md
astroash: ie I would remove "(with timezone?)"

PR #367, opened 2018-04-11
Password hashing

### LINKED ISSUE #213: Passwords being hashed without salt!!
Passwords are being hashed using HMAC with a secret. This is not appropriate for storing passwords, HMAC is intended as an integrity measure. See [this StackExchange question](https://security.stackexchange.com/questions/16809/is-a-hmac-ed-password-is-more-secure-than-a-bcrypt-ed-or-scrypt-ed-password) for a detailed explanation about why you should use salts and hashing algorithms with a variable work factor.

We should switch to [bcrypt](https://www.npmjs.com/package/bcrypt) before launching!

eliascodes:
Fixes #213 

#### Changes
See commits

##### Updating Existing Features
* [x] Have I manually tested my changes?
* [x] Have I added/removed/updated the test suite for this feature appropriately?
* [ ] ~~Have I updated the project documentation and README to reflect my changes?~~
* [x] Do my changes affect steps towards the desired project architecture?


eliascodes:
> When will these routes be rewritten?

ASAP if I can help it. I will prioritise helping you both complete deliverables, but all my other time will go towards the new API and/or data model.

Issue #362, opened 2018-04-10
Exporting as a spreadsheet


RogeredBacon:
As a CB-admin I can export all data for each community business as spreadsheet, also export all data as one spreadsheet:

- [x] Technical spike and designs (2 days) #361 
- [x] Implement design (2days)
- [x] Create export endpoint (3 days)
- [x] Construct single CB xls (3 days)

Semi-seperate deliverable to be completed last:

- [x] As a CB I can export data collected through the app and all the sub-filtered versions of it as csv/excel sheets (2 days)

eliascodes:
Please confirm this is all done, then close 🙏 

Thanks

RogeredBacon:
Job Done

PR #374, opened 2018-04-16
9 - As a CB-admin I can export all data for each community business as spreadsheet, also export all data as one spreadsheet.

### LINKED ISSUE #366: 9.4 - Contruct a CSV of all of the data collected by a CB
9.4 of the deliverables states that we want to be able to export a single CSV of the entirety of a CB's data.

- My question is what will this look like?
- Will it just be tables one after the other?
- Do we want activities also or just visits and users?
- Do we want the tables as they are in the DB or should we replace all ID references with their affiliated data? (activitiesId => Swimming)

RogeredBacon:
Fixes #366 

Related to #362 

#### Changes
Deliverable 9 - As a CB-admin I can export all data for each community business as spreadsheet, also export all data as one spreadsheet.
- [x] I can download a collection of all of the data collected through the app for my CB.

- I've created a new route and query for the visits data csv, and reused an old one for the users csv. I've used fast-csv to create csv strings then packed them into a zip with jszip. The download button is on the Admin Settings page.

##### Adding New Features
* [x] Have I manually tested my new feature?
* [x] Is my new feature sufficiently well tested?
* [x] Have I updated the project documentation and README to reflect my changes?
* [x] Does the new feature conform to the desired project architecture ( I think so, but I get the feeling I'm wrong)


client/src/cb-admin/pages/Settings.js
```diff
@@ -94,6 +101,10 @@ export default class SettingsPage extends React.Component {
       errors: {},
       form: {},
       dropzoneMsg: undefined,
+      users: [],
+      visits: [],
+      visitsString: 'This is visitsString!',
```
eliascodes: What are these?

client/src/cb-admin/pages/Settings.js
```diff
@@ -112,15 +123,69 @@ export default class SettingsPage extends React.Component {
           this.props.history.push('/admin/login');
         }
       });
+
+    Visitors.get(this.props.auth)
+      .then((res) => {
+        this.props.updateAdminToken(res.headers.authorization);
+        this.setState({ users: res.data.result });
+        this.state.users.unshift({
```
eliascodes: _NEVER_ alter the state directly! Also, this doesn't make sense because `setState` is in general asynchronous. You cannot rely on `users` to have the values in `res.data.result` by the time you get to line 131.

client/src/cb-admin/pages/Settings.js
```diff
@@ -112,15 +123,69 @@ export default class SettingsPage extends React.Component {
           this.props.history.push('/admin/login');
         }
       });
+
+    Visitors.get(this.props.auth)
```
eliascodes: Why grab all this data on page load? This data is only required in the user clicks the export button. It increases the load on the server/database unnecessarily.

client/src/cb-admin/pages/Settings.js
```diff
@@ -112,15 +123,69 @@ export default class SettingsPage extends React.Component {
...
+      .then((res) => {
+        this.props.updateAdminToken(res.headers.authorization);
```
eliascodes: Is it necessary to update the token after _every_ api call? Does the server change the token?

server/database/queries/visits_all_export.js
```diff
@@ -0,0 +1,12 @@
+const getAllVisitsQuery = `
```
eliascodes: Isn't this basically identical to `visitors_all` except for the `ORDER BY`? Can you not just use the `visitors_all` endpoint instead and then sort on the client?

client/src/cb-admin/pages/Settings.js
RogeredBacon: users and visits are the data I'm getting back from the db; and the strings are the csv strings generated by fast-csv which jszip uses to generate the files within the zip.

eliascodes: Yeah, but why is there irrelevant default data? Empty strings are more appropriate.

server/database/queries/visits_all_export.js
RogeredBacon: I was going to do this but I needed the visitors name rather than their id as is returned by visitors_all.

eliascodes: is it possible to add the `visitor_name` to the other query without breaking anything?

RogeredBacon: Well, currently we show the users id rather than name on visits data page. Would it make more sense to actually display the name? The only reason I need the name for the csv is for clarity rather than an arbitrary number; which holds true for the page as well. In that case I would be able to use the exact same query.

eliascodes: No, what I mean is, send them both, and just pick what is needed for either page

RogeredBacon: Do you mean delete the values `visitor_id: 5` or `visitor_name: Jeremy` from the array of objects in the route?

eliascodes: No, just send both from the backend, and just don't pick the values that aren't necessary on the client.

client/src/cb-admin/pages/Settings.js
```diff
@@ -147,7 +158,91 @@ export default class SettingsPage extends React.Component {
       errors: {},
       dropzoneMsg: undefined,
     });
-  }
+  };
+
+  createZip = () => {
```
eliascodes: Think about what you're trying to do here. You have three asynchronous operations, 
1. fetch visitors
2. fetch visits
3. construct csv

One or more of those operations have dependencies on the others:
* (1) depends on nothing
* (2) depends on nothing
* (3) depends on (1) and (2)

This means (3) must be scheduled _after_ (1) and (2) are complete.

The way that you've structured your code at the moment means that (3) is scheduled to happen after (2), but (1) happens in parallel to (2):
```
(1) ------------
(2) ---- (3) ---
```
Starting (1) and (2) at the same time is fine because they don't have any mutual dependencies, but (3) needs to wait until they're both finished before starting. How might we accomplish that?

eliascodes: If you're still stuck, look into `Promise.all`

RogeredBacon: Awesome, thank you will do

eliascodes:
How's this coming?

eliascodes:
...and resolving conflicts, obviously.

server/routes/__tests__/export_visits.test.js
```diff
@@ -0,0 +1,49 @@
+const test = require('tape');
+const request = require('supertest');
+const jwt = require('jsonwebtoken');
+const createApp = require('../../app');
+const { getConfig } = require('../../../config');
+
+const config = getConfig(process.env.NODE_ENV);
+
+test('POST /api/cb/export | Correct token', t => {
```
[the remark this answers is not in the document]
RogeredBacon: Hah, you're right. It should check for the correct response now.

Issue #363, opened 2018-04-10
Visitor page pagination


astroash:
- [ ] Update queries to support pagination
- update endpoints to accept page number
- update queries to have offset and limitation
psst initial query/route will need to give total lines of data so front end pagination tabs
- update tests
- [ ] Implement pagination component on client app
- add a container for table which includes pagination tabs. Event listeners on tabs will request data which is passed to table
- [ ] Add search bar to Visitor details page to search for both names and emails
- _brains on pagination right now... will update deets soon_

eliascodes:
To answer your question on gitter, here's what I'm imagining:

A function to which I can pass either:
* a string representing a SQL query, and
* an object with `text` and `values` as keys, representing a SQL query

The function then appends the appropriate `LIMIT`/`OFFSET` statements to the query and returns the new query.

Thoughts?

Issue #297, opened 2018-03-23
Make eslint config stricter


astroash:
I'm up for us synchronising our workspace setup with either:
- everyone using the same prettier config
- no one using prettier and making our eslint config stricter

I'm currently finding myself enabling/disabling prettier dependant on what files I'm working on to avoid unnecessary difs.

RogeredBacon:
We should have done this at the very start

eliascodes:
Agreed. FTR, I don't use prettier. What are the eslint files for if everyone on the project was already using prettier?

astroash:
@eliascodes I'm assuming that they started with airbnb eslint config, and have been using prettier to make it more opinionated.

Does anyone have any strong preferences either way? I am keen for everyone to be doing the same thing as this is currently a pain in the ass.

RogeredBacon:
Yeah you're right, Zooey introduced us to prettier. I'd like to continue using prettier if possible

eliascodes:
OK fine by me

astroash:
I'm up for us all using prettier, but overwriting any annoyances with eslint. This could be an intermediary step whilst we make a dreamy liner config or we could continue this way.

astroash:
It's been decided to create the worlds best eslint config and remove prettier. If any linter rules are added please relate them to this post ❤️

PR #378, opened 2018-04-16
🗃 Update queryBuilder

### LINKED ISSUE #369: SQL Query Builder
https://github.com/TwinePlatform/DataPower/blob/18bfbc70870b47a5c429eb360fd6cb676982ac46/server/shared/models/query_builder.js#L22-L40

I'm still not convinced this is a good idea. I think wrapping simple SQL this way takes away from readability of code, and extending this module to support `INNER JOIN` ect is unnecessary we're effectively writing an extensive module without documentation (why not just use something like [knex](http://knexjs.org/)?). 

If we're refactoring our SQL queries we should start with our most complex ones that need it (ie the mess that is the filter queries), and if we're moving to new way of creating SQL queries then an issue should be raised and it should be discussed.

astroash:
**Keen to get this reviewed/merged to avoid an ever growing dif**

Fixes #369 

Related to #363 

#### Changes
* edit query builder to make it more ridiculous
* change all routes which were using query builder
* fix related tests
* update visitors page to sort/filter using queries rather than frontend

##### Updating Existing Features
* [x] Have I manually tested my changes?
* [x] Have I added/removed/updated the test suite for this feature appropriately?
* [x] Have I updated the project documentation and README to reflect my changes?
* [x] Do my changes affect steps towards the desired project architecture?

### DIFF
--- a/client/src/api/index.js
+++ b/client/src/api/index.js
@@ -38,15 +38,12 @@ export const Activities = {
 
 export const Visitors = {
   get: (tkn, query) => {
-    console.log(query);
-
     if (!query) {
       return axios.get(
         '/api/users/all',
         { headers: { Authorization: tkn } },
       );
     } else if (query.page) {
-      console.log(query.page);
       return axios.get(
         '/api/users/all'
         , { params: {
--- a/client/src/api/index.js
+++ b/client/src/api/index.js
@@ -4,20 +4,12 @@
 import axios from 'axios';
 import { map, head, pathOr, equals, compose } from 'ramda';
 
-
 export const Activities = {
   get: (tkn, { weekday = 'all' } = {}) =>
-    axios.get(
-      `/api/activities/${weekday}`,
-      { headers: { Authorization: tkn } },
-    ),
+    axios.get(`/api/activities/${weekday}`, { headers: { Authorization: tkn } }),
 
   create: (tkn, { name }) =>
-    axios.post(
-      '/api/activity/add',
-      { name },
-      { headers: { Authorization: tkn } },
-    ),
+    axios.post('/api/activity/add', { name }, { headers: { Authorization: tkn } }),
 
   update: (tkn, { id, monday, tuesday, wednesday, thursday, friday, saturday, sunday }) =>
     axios.post(
@@ -27,33 +19,23 @@ export const Activities = {
     ),
 
   delete: (tkn, { id }) =>
-    axios.post(
-      '/api/activity/delete',
-      { id },
-      { headers: { Authorization: tkn } },
-    ),
-
+    axios.post('/api/activity/delete', { id }, { headers: { Authorization: tkn } }),
 };
-
 
 export const Visitors = {
   get: (tkn, query) => {
     if (!query) {
-      return axios.get(
-        '/api/users/all',
-        { headers: { Authorization: tkn } },
-      );
-    } else if (query.page) {
-      return axios.get(
-        '/api/users/all'
-        , { params: {
-          page: query.page,
+      return axios.get('/api/users/all', { headers: { Authorization: tkn } });
+    } else if (query.visitors) {
+      return axios.get('/api/users/all', {
+        params: {
+          offset: query.offset,
           sort: query.sort,
           sex: query.genderFilter,
           age: query.ageFilter,
         },
-        headers: { Authorization: tkn } },
-      );
+        headers: { Authorization: tkn },
+      });
     } else if (query.filter || query.sort) {
       return axios.post(
         '/api/visitors/filtered',
@@ -63,7 +45,6 @@ export const Visitors = {
         },
         { headers: { Authorization: tkn } },
       );
-
     } else if (query.name && query.email) {
       return axios.post(
         '/api/visit/check',
@@ -76,7 +57,6 @@ export const Visitors = {
         },
         { headers: { Authorization: tkn } },
       );
-
     } else if (query.id) {
       return axios.post(
         '/api/user/details',
@@ -85,7 +65,6 @@ export const Visitors = {
         },
         { headers: { Authorization: tkn } },
       );
-
     } else if (query.hash) {
       return axios.post(
         query.asAdmin ? '/api/user/qr' : '/api/user/name-from-scan',
@@ -95,26 +74,16 @@ export const Visitors = {
         { headers: { Authorization: tkn } },
       );
     } else if (query.withVisits) {
-      return axios.post(
-        '/api/visitors/all',
-        {},
-        { headers: { Authorization: tkn } },
-      );
-
+      return axios.post('/api/visitors/all', {}, { headers: { Authorization: tkn } });
     }
 
     return Promise.reject(new Error('Invalid query parameters'));
   },
 
-  create: (tkn, {
-    name,
-    gender,
-    yob,
-    email,
-    phoneNumber,
-    emailContactConsent,
-    smsContactConsent,
-  }) =>
+  create: (
+    tkn,
+    { name, gender, yob, email, phoneNumber, emailContactConsent, smsContactConsent },
+  ) =>
     axios.post(
       '/api/qr/generator',
       {
@@ -129,7 +98,10 @@ export const Visitors = {
       { headers: { Authorization: tkn } },
     ),
 
-  update: (tkn, { id, name, gender, yob, email, phoneNumber }) => // eslint-disable-line
+  update: (
+    tkn,
+    { id, name, gender, yob, email, phoneNumber }, // eslint-disable-line
+  ) =>
     axios.post(
       '/api/user/details/update',
       {
@@ -142,7 +114,7 @@ export const Visitors = {
       { headers: { Authorization: tkn } },
     ),
 
-  delete: () => { },
+  delete: () => {},
 
   email: (tkn, { id }) =>
     axios.post(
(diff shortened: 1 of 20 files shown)


client/src/api/index.js
```diff
@@ -43,6 +43,17 @@ export const Visitors = {
         '/api/users/all',
         { headers: { Authorization: tkn } },
       );
+    } else if (query.page) {
+      return axios.get(
+        '/api/users/all'
+        , { params: {
+          page: query.page,
```
eliascodes: Shouldn't this be called `offset`?

astroash: I'm using `page` to refer to the page number viewed on the clientside, and using `offset` to refer to the lines of data we are offsetting by. In queries offset is calculated by `page` * `limit` - `limit`.

server/shared/models/query_builder.js
```diff
@@ -1,90 +1,189 @@
...
- * @param   {Object}   [where={}] Constraint object
- * @returns {Query}               { text: String, values: Array }
+ * @param   {Options}  options    options object
+ * @returns {QueryObject} queryObject
  */
-const selectQuery = (table, columns, where = {}) => {
-  const base = `SELECT ${columns.join(', ')} FROM ${table}`;
-  const hasWhere = Object.keys(where).length > 0;
+const selectQuery = (table, columns, options) => {
+  const prefix = options.pagination ? `COUNT(*) OVER() AS full_count, ` : ``;
```
eliascodes: why don't you just concat this onto `columns`?

eliascodes:
Also your changes are reducing the coverage, would be nice if that wasn't the case.

server/shared/models/query_builder.js
```diff
@@ -1,90 +1,189 @@
...
+ * @returns {QueryObject}
+ */
+
+const addBetweenClause = curry((options, queryObj) => {
+  const parameter = queryObj.values.length + 1;
+
+  if (options.between) {
+    const betweenClause = `${
+      options.between.column
+    } BETWEEN  $${parameter} AND $${parameter + 1}`;
```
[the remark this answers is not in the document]
eliascodes: No, ignore me, got mixed up

server/shared/models/query_builder.js
```diff
@@ -1,90 +1,189 @@
...
+ * @param   {Options}
+ * @param   {QueryObject}
+ * @returns {QueryObject}
+ */
+
+const addPagination = curry((options, queryObj) => {
+  const parameter = queryObj.values.length + 1;
+
+  if (options.pagination) {
+    const offset =
```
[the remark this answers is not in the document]
astroash: Prettier won't let me. This was actually one of the main reasons I originally stopped pushing for us to use it.

client/src/api/index.js
astroash: Do you think it will cause confusion with this referring to a different offset than that used in sql?

I think my preference would be to send a page and limit in query, then work out the offset on the server, but happy to rename if you don't think this is an issue.

astroash:
@eliascodes 
> your changes are reducing the coverage, would be nice if that wasn't the case.

I can add some more tests 😊

> Is the md doc necessary? The JSDocs and module header should be made sufficient to work out how to use the functions

Personally I'm really not into JSDocs, and don't find them readable.

server/shared/models/query_builder.js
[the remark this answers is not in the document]
eliascodes: Oh, you don't want us to use prettier now? I do think we need to converge, cos I'm getting PR's with a bunch of formatting. I'm happy for us to converge onto just an ESLint config though if you'd prefer. Could you update #297 with your thoughts pls?

server/shared/models/query_builder.js
```diff
@@ -1,90 +1,189 @@
...
+    return {
+      text: `${queryObj.text} LIMIT 10 OFFSET $${parameter}`,
+      values: [...queryObj.values, offset],
+    };
+  }
+  return queryObj;
+});
+
+const addReturning = curry((options, queryObj) => {
+  if (options.returning) {
```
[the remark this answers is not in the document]
eliascodes: OK, then ignore this for now.

eliascodes:
Separate docs files for functions/modules means I have to have two files open instead of one, and two places to maintain when I make changes.

What are you thoughts on this? @RogeredBacon would also be good to get your impressions on this.
```js
/*
 * Query builder utilities
 * 
 * Functions to build SELECT/INSERT/UPDATE queries dynamically
 * 
 * Complex queries including WHERE, BETWEEN, SORT and OFFSET clauses can be 
 * constructed by specifying options, as defined by the `SqlOptions` type definition
 * below.
 */
const foo = require('...');

/**
 * Can be passed directly to `pg.Client.query` or `pg.Pool.query`
 * to execute a query on the database
 *
 * @typedef  SqlQuery
 * @property {String} text   Query text with placeholder values ($n)
 * @property {Any[]}  values Substituted into query in order
 */

/**
 * Used to configure more complex queries.
 * 
 * Note that if `offset` or `limit` are specified, `full_count` is returned 
 * in addition to other columns, describing the total number of records for
 * that query.
 * 
 * @typedef  SqlOptions
 * @property {Object}   [where]        Column-value pairs to match records on
 * @property {Object}   [between]      Substituted into query in order
 * @property {String}   between.column Column for between query
 * @property {Any[]}    between.values Sorted two element array of min/max values
 * @property {String[]} [sort]         Columns by which to sort, in order
 * @property {Number}   [offset=0]     OFFSET value
 * @property {Number}   [limit=10]     LIMIT value
 */

/**
 * Constructs a parameterised select query object compatible with node-postgres
 * 
 * @example 'Select all rows'
 * selectQuery('foo', ['*'], { where: { id: 1 } });
 * 
 * @example 'Select rows with aliases'
 * selectQuery('foo', ['bar AS b', 'c'], { where: { id: 1 } });
 * 
 * @param   {String}     table   DB table name to select from
 * @param   {String[]}   columns Columns to select
 * @param   {SqlOptions} options Extend query with additional clauses
 * @returns {SqlQuery}           node-postgres compatible query object
 */
const selectQuery = (table, columns, options) => {}
```

server/shared/models/query_builder.js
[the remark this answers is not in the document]
astroash: 🙈

client/src/api/index.js
```diff
@@ -4,20 +4,12 @@
...
 import { map, head, pathOr, equals, compose } from 'ramda';
 
-
 export const Activities = {
   get: (tkn, { weekday = 'all' } = {}) =>
-    axios.get(
-      `/api/activities/${weekday}`,
-      { headers: { Authorization: tkn } },
-    ),
+    axios.get(`/api/activities/${weekday}`, { headers: { Authorization: tkn } }),
```
eliascodes: Is this prettier?

server/routes/users_all.js
```diff
@@ -1,10 +1,62 @@
...
+
+const removeProperties = omit(['sort', 'offset', 'age']);
+const removeEmpty = pickBy(identity);
+
+router.get('/', validate(schema), async (req, res, next) => {
+  try {
+    const query = req.query;
+
+    const addCbId = assoc('cb_id', req.auth.cb_id);
+    const queryPipe = pipe(removeProperties, removeEmpty, addCbId);
```
eliascodes: Is `queryPipe` a good name here? Doesn't tell me what it's for. What it's actually doing is constructing the `where` object, right? What about `createWhereOpts` or `createWhereObj` or `createWhereClauseOpts` or something like that.

server/shared/models/query_builder.js
```diff
@@ -1,44 +1,140 @@
...
  * @param   {String}   table      Table name
  * @param   {String[]} columns    Columns to select
- * @param   {Object}   [where={}] Constraint object
- * @returns {Query}               { text: String, values: Array }
+ * @param   {Options}  options    options object
+ * @returns {QueryObject} queryObject
  */
-const selectQuery = (table, columns, where = {}) => {
+const selectQuery = (table, columns, options) => {
+  options.pagination && [`COUNT(*) OVER() AS full_count`].concat(columns); //eslint-disable-line
```
eliascodes: This does nothing.

astroash: 😅, yeah I was struggling to think of the best way to do this with concat over my previous ternary. What do you suggest?

client/src/api/index.js
```diff
@@ -27,22 +19,23 @@ export const Activities = {
...
-
 export const Visitors = {
   get: (tkn, query) => {
     if (!query) {
-      return axios.get(
-        '/api/users/all',
-        { headers: { Authorization: tkn } },
-      );
+      return axios.get('/api/users/all', { headers: { Authorization: tkn } });
+    } else if (query.visitors) {
```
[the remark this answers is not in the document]
astroash: This was my way of hacking it together until we fix this file. `query.withVisits` on line 76 is being use in the `Visits` component I'm also working on at the moment, and it has all the same params but a different endpoint.

server/shared/models/query_builder.js
eliascodes: const cols = (options.pagination ? 'COUNT(*) OVER() AS full_count' : []).concat(columns);
const base = `SELECT ${cols} FROM ${table}`;

RogeredBacon:
> Separate docs files for functions/modules means I have to have two files open instead of one, and two places to maintain when I make changes.
What are you thoughts on this? @RogeredBacon would also be good to get your impressions on this.

I think having more files to maintain is the crux of the issue for me; anything that can keep the file count low and mean less things to think about when implementing something is good in my book. But yeah they are confusing to look at, the only time I've used them previously was at uni.

eliascodes:
Dunno where we landed on the documentation thing, but happy to deal with it later.

astroash:
A suggestion for alternative documenting of functions:

```
/* 
Query = 
  {
    text: String,
    values: Array a
  }

Options = 
  {
    where: String,
    between: { column: String, values: [ Int ] },
    sort: String,
    offset: Int,
    returning: String
  }
*/

// addWhereClause :: Options -> Query -> Query
```

**Pros** - After aliases, type signatures are more concise
**Cons** - No way to obviously show properties in objs are optional (considered `Maybe` but this isn't strictly true), probably not as widely used as jsdocs (but then again neither is ramda).

// [Sounds cool, I wanna know more...](https://github.com/ramda/ramda/wiki/Type-Signatures)

eliascodes:
OK I'm happy to go with that, but would additionally like to suggest:
* we use `[a]` for arrays, not `Array a`
* we don't use curried type annotations on functions that aren't curried
* type declarations go in the module header comment block
* simple type aliases can be declared inline [like here](https://github.com/ramda/ramda/wiki/Type-Signatures#type-aliases)

But we don't need any of this to be done in this PR.

astroash:
as

PR #397, opened 2018-04-30
🔓 get rid of insecure node-base64-image dependency

### LINKED ISSUE #394: Patch security vulnerability in dependency
`Hoek` is insecure < v5.0.3

eliascodes:
Fixes #394 

This took me a lot longer than I thought because I uncovered several bugs in the process 😖 

#### Changes
* Add custom image-encoding function
* Remove insecure dependency
* Update P2C PNG to use standard filtering algorithm
* Fix bug in user details endpoint
* Fix several bugs in `pdfGenerator` (previously undiscovered because of a missing `await` statement when `pdfGenerator` was being called)

##### Fixing Bugs
* [x] Have I added a (set of) test(s) that proves the bug is fixed?

@astroash the bug fix in the details endpoint was to do with the call signature of the DB query was not updated in the handler after your changes (i.e. passing options objects around). This bug may manifest in other handlers, please take a quick look around if you have the time.


eliascodes:
> The png filtering issue is interesting, will this effect what images a CB can upload on cloudinary that will work with the app?

I don't know tbh. Hopefully Cloudinary re-encodes uploaded images into standard forms.

PR #412, opened 2018-05-02
🚑 fix instascan critical error

### LINKED ISSUE #411: Reported: QR Scanning bug
Edward, via the google form:

> Samsung Galaxy Tab E

> I was trying to scan a QR code. It presented an error, which said "Register or new user" or "Try to Scan again", I chose scan again. The device would register no further QR code input until I restarted it. This is Edward 50% reporting this and 50% testing out this error reporter :)

eliascodes:
Fixes #411 

#### Changes
Instascan was repeatedly throwing errors "Uncaught abort() at Error".
Looks like this was due to not having proper clean-up code in the QRcode
component.

Added a `componentWillUnmount` hook to stop the scanner. Also moved the
scanning logic directly into `componentDidMount` instead of splitting it
between two methods, and renamed the `login` state fragment to a more
meaningful `hasScanned` boolean.

##### Fixing Bugs
* [ ] ~~Have I added a (set of) test(s) that proves the bug is fixed?~~ N/A
* [x] Have I documented any necessary developer notes/gotchas/lessons learnt from this bug fix?

### DIFF
--- a/client/src/visitors/pages/qrcode.js
+++ b/client/src/visitors/pages/qrcode.js
@@ -52,8 +52,9 @@ const SnakeContainerRow = FlexContainerRow.extend`
   width: 100%;
   justify-content: space-between;
   &:nth-child(2n) {
-  flex-direction: row-reverse;
-  `;
+    flex-direction: row-reverse;
+  }
+`;
 
 const capitaliseFirstName = name => name.split(' ')[0].replace(/\b\w/g, l => l.toUpperCase());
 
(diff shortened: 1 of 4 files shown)


client/src/visitors/pages/qrcode.js
```diff
@@ -108,54 +124,14 @@ export default class QRCode extends Component {
       });
   }
 
-  componentWillUpdate(nextProps, nextState) {
-    if (nextState.username === 'there is no registered user') {
```
astroash: As we're no longer matching on this string, can this be removed from the route?

https://github.com/TwinePlatform/DataPower/blob/5dd8b49ab457c64245b50a7d5b54d6c70eb1ac88/server/routes/visitor_name.js#L18

Also in the above res it's sending the `hash: 0` property. Its also worth checking that the case this was previously solving, is also resolved in the rewrite.

PR #414, opened 2018-05-02
🐛 Fix error with missing visits data

### LINKED ISSUE #409: 🐛 New entries to database acting weird when exported
From export data it looks like:

- new users date of signup is not a real js date object
- said new users activities are not showing in export

Will 🔍 investigate...

astroash:
Fixes #409 

#### Changes
* changes column of cb_id that we're matching on. I've no idea why this fixes the bug, if anyone can explain that would be great and I'll add it to the docs
* Also changed the order to ascending so the most recent visits are at the bottom

##### Fixing Bugs
* ~~[ ] Have I added a (set of) test(s) that proves the bug is fixed?~~
* ~~[ ] Have I documented any necessary developer notes/gotchas/lessons learnt from this bug fix?~~

### DIFF
--- a/server/routes/visitors_all.js
+++ b/server/routes/visitors_all.js
@@ -27,7 +27,7 @@ const renameActivityAndGender = renameKeys({
 router.get('/', validate(schema), async (req, res, next) => {
   try {
     const { query } = req;
-    const addCbId = assoc('activities.cb_id', req.auth.cb_id);
+    const addCbId = assoc('users.cb_id', req.auth.cb_id);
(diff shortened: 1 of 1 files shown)


RogeredBacon:
Do we want the most recent visits to be at the bottom? It seems counter-intuitive to get the user to scroll/go to the last page to see the most recent visits when they would otherwise see them immediately.

eliascodes:
If I have time I'll set some aside to look at what was going on here.

PR #420, opened 2018-05-11
⛑ Enforce https

### LINKED ISSUE #410: Enforce HTTPS
Should enforce HTTPS connections either using [express-sslify](https://www.npmjs.com/package/express-sslify) or a simple custom written middleware.

astroash:
Fixes #410

#### Changes
* add `express-sslify`

##### Adding New Features
* [ ] Have I manually tested my new feature? 🤷🏽‍♀️ can't test, pipelines aren't setup
~~* [ ] Is my new feature sufficiently well tested?~~
~~* [ ] Have I updated the project documentation and README to reflect my changes?~~
* [x] Does the new feature conform to the desired project architecure?

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -6,7 +6,9 @@ const cookieParser = require('cookie-parser');
 const bodyParser = require('body-parser');
 const postmark = require('postmark');
 const pg = require('pg');
+const enforce = require('express-sslify');
 const { notFound, errorHandler, validationError } = require('./shared/middleware');
+const { PRODUCTION } = require('../config/environments');
 
 
 module.exports = (cfg) => {
@@ -23,6 +25,9 @@ module.exports = (cfg) => {
   app.set('client:postmark', postmarkClient);
   app.set('client:psql', psqlPool);
 
+  if (cfg.env === PRODUCTION) {
+    app.use(enforce.HTTPS({ trustProtoHeader: true }));
+  };
   app.use(favicon(path.join(cfg.root, 'client', 'public', 'favicon.ico')));
   app.use(logger(cfg.env));
   app.use(bodyParser.json());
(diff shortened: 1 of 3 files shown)


server/app.js
```diff
@@ -22,7 +24,11 @@ module.exports = (cfg) => {
   app.set('cfg', cfg);
   app.set('client:postmark', postmarkClient);
   app.set('client:psql', psqlPool);
-
+  app.use((req, res, next) => {
+    cfg.env === PRODUCTION // eslint-disable-line
```
eliascodes: This doesn't make sense. If we're in the production environment, the middleware will:
1. not call `next()`
2. not return anything
3. not send a response

This breaks the server. Take another look at the `sslify` docs.

astroash: > [In case of a non-encrypted HTTP request, express-sslify automatically redirects to an HTTPS address using a 301 permanent redirect.](https://github.com/florianheinemann/express-sslify)

From my understanding of this, anyone who tries to hit any route with a GET request will be redirected to the https address, so will make a new call to the server. I guess I will need to call next after `enforce.HTTPS()` is called, so this is allowed through, but I'm not sure what the need for point 2 & 3 are.

eliascodes: Look at the examples. None of them look like that ☝️. You're just creating a middleware that does nothing.

astroash: So...
```js
if (cfg.env === PRODUCTION) {
   app.use(enforce.HTTPS({ trustProtoHeader: true }));
};
```
Which is weird because my previous code was returning `403` for POST requests in tests, but maybe that was more to do with scope than design 🤷🏽‍♀️

eliascodes:
@astroash incidentally, you can test this locally...just set `NODE_ENV` to production.

PR #421, opened 2018-05-13
🤷🏽‍♀️Production Logs

astroash:
I got completely sidetracked with this...

After reading up on logging libraries, I thought there wasn't much use in these as we're not saving out logs to a db or writing them to files, as Heroku stores our logs for us.

I also don't see much use in these logs being much different to our dev logs, so I've kept them the same (our dev logs aren't that verbose anyway)... Although I did add a little colour coding to make it easier to scan for errors.

Then I tried to test it locally by editing env vars, and nothing works, I dunno what I'm doing 🤷🏽‍♀️


eliascodes:
@astroash your job to review this now

server/logger.js
```diff
@@ -0,0 +1,37 @@
...
+    const color = chalk[getColor(status)] || chalk.reset;
+
+    return [
+      color(tokens.method(req, res).padEnd(7)),
+      color(tokens.url(req, res).padEnd(30)),
+      color(tokens.status(req, res)),
+      reset(`${tokens['response-time'](req, res)}ms`.padStart(8)),
+      reset('-'),
+      reset(tokens['cb-name'](req, res)),
+      reset(tokens['req-body'](req, res)),
```
astroash: I don't think it's a good idea for us to be logging CBs passwords

Issue #448, opened 2018-07-25
Test react components that handle ajax requests


astroash:
Create tests for react components that handle requests to ensure that the DOM is correctly updated on success and failure.

eliascodes:
Pretty sure the first pass of this was done a while ago.

PR #449, opened 2018-07-25
React testing

astroash:
Related to #448

#### Changes
- fix App layout post removal of prettier
- add component tests for homepage
- move error message outside of label (element cannot be grabbed for testing from inside)
- create renderWithRouter helper to allow testing with faux history object
- test how component responds to successful & unsuccessful statuscodes


client/src/cb-admin/pages/Login.js
```diff
@@ -40,19 +39,20 @@ export default class Login extends React.Component {
 
     CbAdmin.login({ email: this.state.email, password: this.state.password })
       .then((res) => {
-        localStorage.setItem('token', res.data.result.token);
+        if (process.env.NODE_ENV !== 'test') localStorage.setItem('token', res.data.result.token);
```
eliascodes: Is this stripped out by babel at build time? `process` is a Node API and will cause an exception on the browser.

astroash: This was working fine locally, (ie no errors and you can actually console.log(process.env) in the browser), but I guess this is something to do with CRA's hot reloading. Any suggestions of a better approach? if not I could probs do:
```
if(process && process...)
```

eliascodes:
Oh man, such a big hit to coverage 😓

astroash:
> Oh man, such a big hit to coverage 😓

@eliascodes technically, we've doubled the codebase being monitored for coverage, and have _just about_ dropped by less than half...

I see this as a success 🎉

.travis.yml
```diff
@@ -28,6 +28,8 @@ script:
   - npm run build.client
   - npm run cover
   - npm run cover.report.ci
+  - npm run test:client
+  - npm run cover:client
```
eliascodes: Why run both test and cover?

client/src/cb-admin/pages/__tests__/Login.test.js
```diff
@@ -0,0 +1,57 @@
...
+import MockAdapter from 'axios-mock-adapter';
+import renderWithRouter from '../../../tests';
+import Login from '../Login';
+
+
+describe('Login Component', () => {
+  afterEach(cleanup);
+
+  test(':: incorrect user details returns 401 and displays error message', async () => {
+    const mock = new MockAdapter(axios);
```
eliascodes: Do we need to create a new instance of the mock adapter for each test case? Can we just re-use it?

Issue #450, opened 2018-07-26
Adding base url to jsdom tests


astroash:
A base url can be added to jsdom via jest config:
```
"testURL": "http://localhost:4000",
```

However, CRAp only allows a small set of jest configurations, this not being one of them. If we want to add a base url to test react with a live api we will need to eject CRAp.

I'm pro ejecting, but not sure if there are any unintended drawbacks of this.

eliascodes:
Are you pro-ejecting for any other reasons?

eliascodes:
I guess I would prefer to have fewer things to maintain, and staying jacked in to CRA means we can pull updates if/when they come. If we use the package you linked, looks like we'd just need one additional file, right?

astroash:
Rewired allows you to edit jest config from package.json, so we would only need to add an empty config file. I had a go at setting it up here #453. Currently a number of tests fail if ran against the live server due to auth tokens, but can look into this further if we go down this route.

astroash:
in reference to #453:

@eliascodes apologies if it wasn't clear which parts of the diff are related to installing and setting up `react-app-rewired`. Most of your questions are related to the test mock which are a part of other PRs. I set this branch up as an example of how we could run react tests against our new api. Currently I think we have two choices moving forward:

**1️⃣ Setup rewired** to run tests against our new api. This will involve:
- creating mock auth tokens which can be used with our server
- having functionality to turn off mocks in our tests
- using a module such as [`node-vcr`](https://www.npmjs.com/package/node-vcr) so we don't need to also complete the second option.

**2️⃣ Create mocks for new api** manually. This will involve:
- writing mocks for each react test that accepts and returns data in the new structure from the new endpoints
- having functionality to switch between the old mocks and new mocks in our tests

eliascodes:
Prefer option 1 tbh, since option 2 doesn't really sound optional, more like "either we do this manually or automatically". Looks like `node-vcr` is based on [`yakbak`](https://github.com/flickr/yakbak) which IMO seems like a better bet.

> having functionality to turn off mocks in our tests

What does this mean? That we need a way to regenerate whatever data files are created by VCR-like packages?

astroash:
> > having functionality to turn off mocks in our tests

>What does this mean?

Currently I have manually created mocks for react tests. Turning them "off" would involve allowing our tests to run against the live api. This is what I demo's in the `rewired` pr with an `LIVE_API` env variable.

[...]

PR #453, opened 2018-07-27
Rewired Test

astroash:


client/src/cb-admin/pages/__tests__/Activities.test.js
```diff
@@ -15,79 +15,81 @@ import Activities from '../Activities';
...
-          friday: true,
-          saturday: false,
-          sunday: false,
-        },
-        { id: 13,
-          name: 'Skating',
+  if (!process.env.LIVE_API) {
+    beforeAll(() => {
+      mock = new MockAdapter(axios);
+      mock.onGet('/api/activities/all')
```
eliascodes: IMO, actually mocking responses belongs in the test specification itself.

client/src/cb-admin/pages/__tests__/Activities.test.js
```diff
@@ -15,79 +15,81 @@ import Activities from '../Activities';
...
+            name: 'Yoga',
+            monday: false,
+            tuesday: false,
+            wednesday: true,
+            thursday: true,
+            friday: true,
+            saturday: false,
+            sunday: false,
+          },
+          { id: 13,
```
eliascodes: Is this temporary data based on the resolution of #450 and #453? Either way, I think we should put this data into a separate file(s) so as not to bloat the test specs themselves.

client/src/cb-admin/pages/__tests__/Activities.test.js
```diff
@@ -15,79 +15,81 @@ import Activities from '../Activities';
...
-          tuesday: false,
-          wednesday: true,
-          thursday: true,
-          friday: true,
-          saturday: false,
-          sunday: false,
-        },
-        { id: 13,
-          name: 'Skating',
+  if (!process.env.LIVE_API) {
```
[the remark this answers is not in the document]
astroash: This is designed to not be defined if not needed. It's copying the pattern CRA uses to [define jest configs](https://github.com/facebook/create-react-app/blob/90c908e39707d5dc4bf5facdf59bbf755d051f10/packages/react-scripts/scripts/test.js#L36-L43).

client/src/cb-admin/pages/__tests__/ConfirmPassword.test.js
```diff
@@ -14,17 +14,21 @@ import ConfirmPassword from '../ConfirmPassword';
 describe('ConfirmPassword Component', () => {
   let mock;
 
-  beforeAll(() => {
-    mock = new MockAdapter(axios);
-  });
+  if (!process.env.LIVE_API) {
```
eliascodes: If this goes ahead, we'll be doing this quite a lot, so it's worth coming up with a way to accommodate it without having to stick `if` statements all over the place. Maybe something like a function that constructs a new `MockAdapter` instance if `LIVE_API` is `true`, otherwise an object with the same API, whose methods all do nothing?

astroash:
@eliascodes apologies if it wasn't clear which parts of the diff are related to installing and setting up `react-app-rewired`. Most of your questions are related to the test mock which are a part of other PRs. I set this branch up as an example of how we could run react tests against our new api. Currently I think we have two choices moving forward:

**1️⃣ Setup rewired** to run tests against our new api. This will involve:
- creating mock auth tokens which can be used with our server
- having functionality to turn off mocks in our tests
- using a module such as [`node-vcr`](https://www.npmjs.com/package/node-vcr) so we don't need to also complete the second option.

**2️⃣ Create mocks for new api** manually. This will involve:
- writing mocks for each react test that accepts and returns data in the new structure from the new endpoints
- having functionality to switch between the old mocks and new mocks in our tests

PR #475, opened 2018-08-23
🌳 cb admin dashboard

astroash:
- update cbAdmin.get
- remove deprecated route
- fix in components using this method
- create an axios instance for shared config

#473 <- kinda

### DIFF
--- a/client/src/api/index.js
+++ b/client/src/api/index.js
@@ -11,7 +11,7 @@ require('env2')(`${__dirname}/../../../config/config.env`);
 
 const API_HOST = process.env.API_HOST_DOMAIN;
 
-const axios = create({
+export const axios = create({
   baseURL: 'http://localhost:4000/api/v1/',
   withCredentials: true,
 });
--- a/client/src/api/index.js
+++ b/client/src/api/index.js
@@ -12,7 +12,7 @@ require('env2')(`${__dirname}/../../../config/config.env`);
 const API_HOST = process.env.API_HOST_DOMAIN;
 
 export const axios = create({
-  baseURL: 'http://localhost:4000/api/v1/',
+  baseURL: 'http://localhost:4000/v1/',
   withCredentials: true,
 });
 
--- a/client/src/api/index.js
+++ b/client/src/api/index.js
@@ -1,38 +1,45 @@
 /*
  * Twine API interface
  */
-import axios from 'axios';
+import _axios from 'axios'; // eslint-disable-line
+// 👆🏽 to be removed
+
+import { create } from 'axios'; // eslint-disable-line
 import { map, head, pathOr, equals, compose } from 'ramda';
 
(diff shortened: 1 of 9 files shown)


client/src/api/index.js
```diff
@@ -1,38 +1,45 @@
...
+
+import { create } from 'axios'; // eslint-disable-line
 import { map, head, pathOr, equals, compose } from 'ramda';
 
 require('env2')(`${__dirname}/../../../config/config.env`);
 
 const API_HOST = process.env.API_HOST_DOMAIN;
 
+export const axios = create({
+  baseURL: 'http://localhost:4000/api/v1/',
```
eliascodes: Think we need to remove the `/api` from the base URL now?

client/src/cb-admin/pages/Settings.js
```diff
@@ -157,11 +157,11 @@ export default class SettingsPage extends React.Component {
   updateStateFromApi = (data) => {
     this.setState({
       id: data.id,
-      orgName: data.org_name,
-      sector: data.genre,
-      email: data.email,
+      orgName: data.name,
+      sector: data.sector,
+      email: data.email, // not returned in response
```
eliascodes: The CB email isn't in the response from the API? Is that because you're hitting `/organisations/me` instead of `/community-businesses/me`?

astroash: @eliascodes according to the `api.json` community-business is an alias for organisations. If this is no longer the case can you please make an issue so I know what changes are being made.

eliascodes: Sorry, I forgot that `email` is part of the user model now and will need a separate call. Your response stands, we'll talk about that elsewhere.

astroash:
When TwinePlatform/twine-api#118 is merged I'll update organisations endpoints in this PR

client/src/cb-admin/pages/Settings.js
eliascodes: So are we going to do the call to get the `email` in a different PR or in this one?

astroash: @eliascodes are we enforcing one admin account per community business? If so, this can be dealt with in this PR, if not we need to have a think about how we are relating email address to cbs.

eliascodes: > are we enforcing one admin account per community business?

For now, I think so

astroash: If we're enforcing 1 admin account per cb, do we want to add the email address to the cb model?

Or hit the `/community-business/me/users/admins` endpoint?

astroash: Yeah, that makes sense. Technically, then we do not need to enforce 1 admin per cb.

Issue #447, opened 2018-07-24
Integrating with new api


astroash:
- write integration tests for components
- create new api layer like `api/index.js` that builds queries, makes api calls
- create a layer to manipulate response data to be usable by component (to be stored either in component or in `/api/` under component specific files)
- update component to use new api layer
- check tests still pass ✨

PR #512, opened 2018-09-18
Disable cb signup

### LINKED ISSUE #494: Disable CB sign-up page
Due to pending decision in TwinePlatform/twine-api#97, disable CB sign-up page for now.

### LINKED ISSUE #510: 🌳 Update password reset routes
Still pointing at the old API.

Related to #447

eliascodes:
Fixes #494 
Fixes #510

Related to #447

#### Changes
* See commits

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -18,7 +18,6 @@ import Container from './shared/components/Container';
 import HomePage from './shared/pages/Home';
 
 import Login from './cb-admin/pages/Login';
-import CbSignupPage from './cb-admin/pages/Signup';
 import ConfirmPassword from './cb-admin/pages/ConfirmPassword';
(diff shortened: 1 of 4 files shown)


client/src/App.js
```diff
@@ -18,7 +18,7 @@ import Container from './shared/components/Container';
 import HomePage from './shared/pages/Home';
 
 import Login from './cb-admin/pages/Login';
-import CbSignupPage from './cb-admin/pages/Signup';
+// import CbSignupPage from './cb-admin/pages/Signup';
```
astroash: 🔥

eliascodes: Yeah I wasn't sure. Think we should just delete it? Don't mind.

astroash: Yeah, I would. Otherwise it's just a bit of clutter we'll need to remove later.

PR #514, opened 2018-09-18
🔧 add api url base as env variable

### LINKED ISSUE #504: Add support for environments via CRA
https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-development-environment-variables-in-env

This is primarily to allow dynamically changing the API URL, currently hardcoded and preventing any review/staging apps from working.

astroash:
- add api url base as env variable
- NB clientside env vars not to be used for secrets so removed from .gitignore

Fixes #504

### DIFF
--- a/client/.env
+++ b/client/.env
@@ -1,0 +1,1 @@
+REACT_APP_HOST_API=https://api.twine-together.com
--- a/.gitignore
+++ b/.gitignore
@@ -55,7 +55,6 @@ typings/
 .yarn-integrity
 
 # dotenv environment variables file
-.env
 config.env
 
 # OS generated files #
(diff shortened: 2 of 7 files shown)


client/.env.local
```diff
@@ -0,0 +1 @@
+REACT_APP_HOST_API=http://localhost:4000/v1/
```
eliascodes: Check the [precedence rules](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#what-other-env-files-can-be-used), this will always overwrite the other one.

client/.env
```diff
@@ -0,0 +1 @@
+REACT_APP_HOST_API=https://api.twinetogether.com
```
eliascodes: Wrong URL

client/.env.local
astroash: 🤦🏽‍♀️ i over confidently read that the wrong way round

Issue #516, opened 2018-09-18
🐛 visitor sign up route returns 403


astroash:
`users/register/visitor` route returns 403 and forwards to error page when you try to sign up a new user

eliascodes:
Think this can be closed now, yes?

astroash:
yup, looks like my bold **fixes** is breaking auto closing

PR #517, opened 2018-09-18
🐛 send orgId when signing up new visitor

astroash:
**Changes**: send orgId when signing up new visitor

**Fixes** #516

NB: route is still not working due to TwinePlatform/twine-api#159

### DIFF
--- a/client/src/visitors/pages/main.js
+++ b/client/src/visitors/pages/main.js
@@ -117,7 +117,7 @@ export default class Main extends Component {
   }
 
   componentDidMount() {
-    CommunityBusiness.get({ fields: ['name', 'logoUrl'] })
+    CommunityBusiness.get({ fields: ['name', 'logoUrl', 'id'] })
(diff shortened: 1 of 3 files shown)


client/src/visitors/pages/main.js
```diff
@@ -121,6 +122,7 @@ export default class Main extends Component {
         this.setState({
           cbOrgName: res.data.result.name,
           cbLogoUrl: res.data.result.logoUrl,
+          organisationId: res.data.result.id,
```
eliascodes: This wont exist, look at the `fields`.

eliascodes: Also pls update the test payload

astroash: Weird that it works though huh? 🙃

eliascodes: Lol, ok I'll make an issue in `twine-api` then.

PR #520, opened 2018-09-20
🐛 fix cb password reset validation issues

### LINKED ISSUE #519: 🐛 Password reset error
User tried to reset their password but got multiple errors
- [no special char] too weak with no explanation why
- [with !] token has expired
- [with ?] too weak

astroash:
Changes:
- add more special characters to regex
- extend token lifespan to 1 day
- clarify error messages
- add password strength critera on client

Fixes #519

### DIFF
--- a/server/routes/cbauthentication/cb_password_change.js
+++ b/server/routes/cbauthentication/cb_password_change.js
@@ -13,7 +13,7 @@ const schemas = {
     token: Joi.string().required(),
 
     password: Joi.string()
-      .regex(/^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!@#$%^&*])(?=.{8,})/, 'strong_pwd')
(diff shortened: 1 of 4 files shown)


server/routes/cbauthentication/cb_password_change.js
```diff
@@ -13,7 +13,7 @@ const schemas = {
     token: Joi.string().required(),
 
     password: Joi.string()
-      .regex(/^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!@#$%^&*])(?=.{8,})/, 'strong_pwd')
+      .regex(/^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!#$%&()*+,-./:;<=>?@[\]^_{|}~?])(?=.{8,})/, 'strong_pwd')
```
eliascodes: Do `$`, `^`, `?`, `.`, `*`, need to be escaped? Also you've `?` twice.

astroash: @eliascodes quantifiers don't work in character sets, but I have missed a few escapes.

PR #528, opened 2018-10-11
Fix 101

astroash:
Related to TwinePlatform/twine-api#190

#### Changes
See commits

### DIFF
--- a/client/src/cb_admin/pages/Settings.js
+++ b/client/src/cb_admin/pages/Settings.js
@@ -110,9 +110,9 @@ export default class SettingsPage extends React.Component {
 
   componentDidMount() {
     const { get } = CbAdmin;
-    const { get: getCb, sectors, regions } = CommunityBusiness;
+    const { get: getCb, sectors } = CommunityBusiness;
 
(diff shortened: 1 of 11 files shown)


client/src/cb_admin/pages/Signup.js
```diff
@@ -24,6 +24,7 @@ import {
   ORG_NAME_INVALID,
 } from '../constants/error_text';
 
+// TODO: get from api on mount
```
eliascodes: Should we do this now, in case we forget later?

astroash: Ya, can do

client/src/cb_admin/pages/Settings.js
```diff
@@ -113,11 +113,10 @@ export default class SettingsPage extends React.Component {
     const { get: getCb, sectors, regions } = CommunityBusiness;
 
     Promise.all([get({ fields: ['email'] }), getCb(), sectors(), regions()])
-      .then(([resUser, resCb, rSectors, rRegions]) => {
+      .then(([resUser, resCb, rSectors]) => {
```
eliascodes: if we don't need the result, we can also remove the API call on line 115?

PR #548, opened 2018-11-06
🐛 Fix homepage && password pages

astroash:
**Changes:**
- Homepage makes api call to redirect to login if no cookie is present
- get base url from node environment
- redirect to correct page for qr code errors
- update pw forgot/reset to work with new api
- add redirect payload to forgot
- add email to reset

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -5,9 +5,12 @@ import _axios, { create } from 'axios';
 import { pathOr, equals, compose } from 'ramda';
 import qs from 'qs';
 
+const baseURL = process.env && process.env.NODE_ENV === 'development'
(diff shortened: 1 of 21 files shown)


src/api/index.js
```diff
@@ -5,9 +5,12 @@ import _axios, { create } from 'axios';
 import { pathOr, equals, compose } from 'ramda';
 import qs from 'qs';
 
+const baseURL = process.env && process.env.NODE_ENV === 'development'
```
eliascodes: I feel like you mentioned something about this but don't remember why we've gone back to doing this instead of using `REACT_APP_` env vars?

PR #554, opened 2018-11-13
🐛 Update instascan module

### LINKED ISSUE #547: ⚠️ instacam using deprecated api
instacam uses URL.createObjectURL which will be deprecated from December 2018. There is a pr waiting to fix this on their repo:

see https://github.com/schmich/instascan/issues/177

supersedes #545

astroash:
**Changes:**
- add updated instascan.min.js
- add notes on how to rebuild
- fix 💩 code

Fixes #547


src/visitors/pages/qrcode.js
```diff
@@ -101,6 +99,10 @@ export default class QRCode extends Component {
 
         Visitors.search({ qrCode: content })
           .then((res) => {
+            if (!res.data.result) {
+              throw new Error('No user found');
```
eliascodes: User gets no feedback if we throw.

astroash: Falls into the catch, which is the qr error page. I don’t think we need any additional feedback (at this point)

astroash: Falls into the catch, which forward to the qr error page. I don’t think we need any additional feedback (at the moment)

eliascodes:
The review app crashes: https://twine-visitor-pr-554.herokuapp.com

Is this because we've mis-configured it?

astroash:
@eliascodes fixed the heroku build issues, but you will not be able to log into the app until the staging branch db is rebuilt with the changes on TwinePlatform/twine-api#229

eliascodes:
The app is being run in development mode, not production mode:
screen shot 2018-11-13 at 18 01 03

astroash:
@eliascodes review app is now up

eliascodes:
Why is 4b60c08 necessary? We were using the `create-react-app-buildpack` #544

eliascodes:
I get 404 error on the app

PR #573, opened 2018-12-07
🚸 improve UX of QR error page by passing error message in qs

eliascodes:
### Changes
* Make buttons consistent with rest of app's style
* Display error message if passed through the query string
* If no user found, display a message saying so


astroash:
Merged the other pr instead

Issue #594, opened 2018-12-17
App crash on sign in


eliascodes:
Reported by @edwardwptc via e-mail:

> ...our most regular visitor app user, are reporting that they are logged in successfully but pressing "Sign in with QR code" takes them to a completely blank white screen. 

A completely blank screen suggests an unhandled exception that causes React itself to no longer be able to render a page.

So this cannot be reproduced locally (normally), except with the following steps:
1. Log in
1. Before going to the visitor sign in page, set `window.Instascan = undefined` in the console
1. Click the sign in button

If the absence of the `Instascan` object is in fact the reason for the failure, it suggests something strange is going on with how that script is being loaded. Normally something like a page refresh should help, however:

> ...they are still having the issue. They have refreshed the browser and fully restarted the tablets.

That doesn't seem to have helped in this case. If it weren't for the fact that the service worker were unregistered in #592, I would suspect that it might have something to do with that. Or perhaps the browser cache.

edwardwptc:
They report still seeing a white screen, no refresh button. v.71.0.3578.98

eliascodes:
Have cloned the production database locally in order to log in as the account in question, and was unable to reproduce. This rules out a crash due to API issues or data inconsistencies.

astroash:
user has not relogged in since launch

PR #595, opened 2018-12-17
prevent crash in absence of Instascan

eliascodes:
### Changes
* Attempts to address #594 by detecting if `Instascan` is available


eliascodes:
Self merging(!)

Issue #536, opened 2018-10-17
Visitor app: sign in with name allowed


SonjaTwine:
add option next to sign in with qr code
no auto filling or name suggestions! type full name
produces an error if not existing visitor

edwardwptc:
Okay, that looks like an excellent way to move forward with this. Very much like the idea of separating out the "created" people and the anonymous people as different creation functions. Makes it much more logical. 

Questions:

"Sign in flow
 if a user does not have a qr code with then they can select sign in with name
 if their name is a duplicate then the server will request the next piece of data we have on the user (ie if they have not supplied DOB, then we will go to postcode ect)"

Didn't you say that:
"Unique users:

    fill out sign up form
    they either need to supply an email OR a phone number"

So won't "if name is a duplicate" resolve it via the email or phone number we are requiring? 

My preference would definitely be to find some way of allowing names with extremely minimal data, but in the interests of time (and the fact our future effort to allow admins to customize mandatory signup data will inherently require us to address these questions at that stage) this looks like a really solid solution.

Very much like the "Sign in Anonymously" button idea, with a drop-down then of whatever the CB created, i.e. Anonymous Male, Anonymous Female, Anonymous Child. etc. It's not ideally what PTC or Gen originally visualised with the personal data but it absolutely is how people want to use it.

astroash:
Great, good to hear this approach is more in line with how cbs want to use the app!

> So won't "if name is a duplicate" resolve it via the email or phone number we are requiring?

Good spot, we can do this. We already part of the sign in with DOB/post code written so if think this would be more useful as it requires less data entry we can use this approach.

edwardwptc:
I wasn't necessarily proposing a change, just looking to understand why we'd require data sets for unique identification and then not validate with those data sets. Because obviously some CB's will find that data in some way or other, a hassle, to *have* to capture it.

If it's the best current route towards a functioning name sign up, I'm all for it. 100% 

We will want to review this in future as part of the process of allowing CB's the ability to customise signup data. Making an issue for that one shortly.

eliascodes:
@astroash can we just close this for now and re-open once we get feedback on the feature? Otherwise it'll be hanging around "in-progress" for months.

astroash:
Can it be moved to backlog?

PR #600, opened 2018-12-21
Sign in w name

### LINKED ISSUE #508: 💅🏻 No styling for buttons on qr error page
screen shot 2018-09-18 at 1 21 34 pm

### LINKED ISSUE #562: 🐛 Cannot differentiate between users with the same name
Visitor login with name does not differentiate between users with the same name.

I wonder if this feature would be better implemented with email or matching names with year of birth

eliascodes:
Related to #536
Fixes #562
Fixes #508 

Fairly big PR because the changes required for #562 meant it didn't make any sense to continue with the old structure.

#### Changes
* Install flexbox grid b/c how the hell did we not have a grid system before
* Heavy refactor of the visitor sign in page, splitting out several sub-components
* Allow signing in with name, falling back first to birth year, and then to e-mail if several users with the same name (or name + birth year) are found.
* Update the routing to:
   1. Make routes consistent (e.g. get rid of `/admin/*` routes)
   2. Move Visitor routes into a `VisitorRoutes` component and likewise with CB admin routes (which better reflects the intended [project architecture](https://github.com/TwinePlatform/twine-visitor/blob/staging/docs/architecture.md#general)
* Update docs + get rid of (most of) PR template.

#### Notes
* The styling changes made to `LabelledInput` and `LabelledSelect` will affect other pages but they've not been "fixed" in this PR because it's already too big. Also the weird 90% width thing I set them too was a bad idea anyway.


src/shared/constants.js
```diff
@@ -0,0 +1,89 @@
...
+  toAge: year =>
+    year === null ? null : moment().year() - year,
+
+  // fromAge :: Number | null -> Number | null
+  fromAge: age =>
+    age === null ? null : moment().year() - age,
+
+  // toDisplay :: Number | null -> String
+  toDisplay: year =>
+    year === null ? 'Prefer not to say' : year,
```
astroash: I think it should be the responsibility of the server to do this, not the client.

src/visitors/pages/qrerror.js
```diff
@@ -24,7 +25,7 @@ const ContinueOptions = {
         <Button>Register as a new user</Button>
       </Link>
       <Link to="/visitor/login">
-        <Button>Try to scan QR code again</Button>
+        <Button>Try to sign in code again</Button>
```
astroash: Wording doesn't make sense

src/shared/constants.js
eliascodes: IMO the proper place for presentational logic is the client/frontend code.

astroash: My preference for this being handled by the server is due to how filtering and aggregating data will look.

- requests can be filtered by age. If this is handled by the client then queries will be age=null
- aggregated data is handled by the server and will return data for age group 'prefer not to say'.

astroash:
From user testing:

- no option for no linked YOB. Prefer not to say should be in the dropdown (not sure how good this is for ux)

src/shared/constants.js
eliascodes: > aggregated data is handled by the server and will return data for age group 'prefer not to say'.

I may be missing a piece here, so forgive me.

Are we talking exclusively about `/community-bussinesses/me/visit-logs/aggregates` and `/community-businesses/me/visitors`?

The age filter is currently implemented as a two-element array of numbers. This will need to change in any case, since it doesn't support the new use case.

I don't see why we would prefer `?age=prefer not to say` instead of `?age=null`. I know I said differently [here](https://github.com/TwinePlatform/twine-visitor/issues/584#issuecomment-448624123), but that was just my first reaction. I still think ->

> IMO the proper place for presentational logic is the client/frontend code.

eliascodes:
> This may cause ux issues as users must remember if they specified this on sign up.

Yeah this is probably a fair point. Not clear what the best option would be though, so not sure I want to do anything about it right now (also everyone already signed up will have a YOB).

astroash:
@eliascodes heroku pipeline for this branch is forwarding to 500 error page on load. Do you know if this is something that has changed in this PR? This is not happening on the staging pipline.

eliascodes:
@astroash need to configure proxies on review apps I think

astroash:
@eliascodes yeah, weirdly heroku has stopped pulling env vars from staging. Fixed now.

I'm keen to wait for this to be merged, I would like #603 to be merged first, which is waiting on TwinePlatform/twine-api#289

eliascodes:
Why does #603 need to go in first? There's a lot of work behind TwinePlatform/twine-api#289

astroash:
This PR is a complete restructure and wasn't mentioned. It changes a number of files that are edited in my PR which is branched off of staging.

src/visitors/pages/Login.js
```diff
@@ -0,0 +1,272 @@
...
+        err => redirectOnError(this.props.history.push, err, { default: '/visitor/qrerror' }),
+      )
+
+  submitVisitorName = ({ name }) =>
+    this.searchVisitorByDetails({ name })
+
+  submitVisitorBirthYear = ({ name, birthYear }) =>
+    this.searchVisitorByDetails({
+      name,
+      age: birthYear ? [1, 1].map(() => moment().year() - birthYear) : undefined,
```
eliascodes: THink about whether this makes sense as part of BirthYear/Age module, or think about adding birthYear as filter option.

src/visitors/pages/Login.js
```diff
@@ -0,0 +1,272 @@
...
+
+  onCameraError = () => {
+    this.props.history.push('/visitor/qrerror?e=camera');
+  }
+
+  onScannerError = () => {
+    this.props.history.push('/visitor/qrerror?e=scanner');
+  }
+
+  onSignInSuccess = (res) => {
```
eliascodes: Not explicit what's happening here, at least destructure arguments for more clarity.

src/tests/mock_responses.js
```diff
@@ -0,0 +1,37 @@
+export const Activities = {
+  get: {
+    200: (mock, params, response) =>
+      mock.onGet('/community-businesses/me/visit-activities', { params }).reply(200, response || { result: [] }),
```
eliascodes: Lot of repetition here, think about reducing.

astroash:
@eliascodes does this feature allow searching for numbers that are separated by spaces?

eg if I type `000 000 000` will it match `000000000`?

eliascodes:
Do you mean in the "name" field? If so, no, it will read the value entered as a string, and so preserve spaces.

astroash:
I meant phone numbers as we were talking about this being a potential value to identify users. On second look this feature isn't currently matching on phone number values. Worth keeping in mind if we do end up implementing this.

Also, I'm keen to revert these commits as it looks like this feature may not be ready for a while, and this PR is incomplete.

astroash:
@eliascodes if the restructure commits are entirely separate we can keep those and only revert the feature commits.

PR #601, opened 2018-12-21
🐛 fix proxy issues

astroash:
#### Changes
- build base url dependant on wether HOST_DOMAIN is set
- remove HOST_DOMAIN from required env vars in heroku config

Heroku changes:
- added `PROXY_API_URL` to staging and PR pipelines for test apps. Previously experiencing cors issues

@eliascodes with the removed heroku config, we may want to throw an error on start if either `PROXY_API_URL` or `REACT_APP_API_HOST_DOMAIN` isn't set

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,11 +1,6 @@
 {
   "name": "twine-visitor",
   "scripts": {},
-  "env": {
-    "REACT_APP_API_HOST_DOMAIN": {
-      "required": true
-    }
-  },
   "formation": {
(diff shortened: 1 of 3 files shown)


astroash:
Self merging

astroash:
@eliascodes currently the version number is appended in the axios base url, so all ajax requests are going to `.../v1/v1/...`

Moving the static serving before this should fix that problem.

PR #618, opened 2019-01-10
Fix activities page 403

### LINKED ISSUE #617: 403 response on activity page gives poor feedback
Inspecting the API logs recently I found that at `Jan 10 12:58:07`, a number of 403 responses were sent to the visitor app while a user was trying to update an activity. This is because their escalated session had expired, and they needed to log-in/escalate again. However, the existing component gave no such feedback (merely a vauge "could not update" message).

This should be fixed.

eliascodes:
**MERGE #615 FIRST**

Fixes #617

#### Changes
* Use `redirectOnError` on Activities page


eliascodes:
Looks like the conflicts disappeared 🤷🏼‍♂️

PR #628, opened 2019-01-15
15-01-19 Deployment

eliascodes:
Additional user flows to test:
* ...


astroash:
From user testing:
- redirect without escalated cookie only works if the page has a on mount request. Navigating to `/cb/dashboard` does not cause a redirect.

PR #633, opened 2019-01-15
Fix forgot pw validation disp

### LINKED ISSUE #632: ForgotPassword validation error not being displayed
This is as a result of the `/password/forgot` route now (IMO correctly) returning `400` in the case of an unrecognised e-mail, rather than a 401.

Minor, easy fix.

eliascodes:
Fixes #632 

### Changes
* Add `ErrorUtils.getErrorMessage` utility function
  * Motivated by cases where response is rightly `400`, but isn't generated by `Joi` (and thus the `validation` property is not set)
* Remove `401` branch from `ForgotPassword` since it's impossible
* Add missing `403` response branch to `ResetPassword` (happens if email doesn't correspond to real user)
* Update status code in test to reflect API

### User flows to test

- [x] * Successful forgot/reset password flow
  - I've changed the url in postmark email template as one was missing email param
- [x] * Failed reset password because of bad token
- [x] * Failed reset password because of non-existent e-mail
- [x] * Failed forgot password because of non-existent e-mail

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -2,7 +2,7 @@
  * Twine API interface
  */
 import _axios, { create } from 'axios';
-import { pathOr, equals, compose, evolve, map } from 'ramda';
+import { pathOr, equals, evolve, map } from 'ramda';
 import qs from 'qs';
 import { BirthYear } from '../shared/constants';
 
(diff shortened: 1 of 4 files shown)


astroash:
@eliascodes in future, can you fill in the description with details of changes. It makes it easier to review.

PR #638, opened 2019-01-16
🐛 fix 404 routing for nested routes

### LINKED ISSUE #637: 🐛 404 not working for nested routes
Going to `/cb/does-not-exist` or `visitor/whereami` does not result in a 404 due to clientside routing changes

astroash:
Fixes #637 

#### Changes
* nested routes use switch statement so they can return 404 if no matching page is found

#### User testing flows:

- [x] `/cb/does-not-exist` returns 404 page
- [x] `/visitor/does-not-exist` returns 404 page

### DIFF
--- a/src/visitors/index.js
+++ b/src/visitors/index.js
@@ -1,5 +1,5 @@
-import React, { Fragment } from 'react';
-import { Route } from 'react-router-dom';
+import React from 'react';
+import { Switch, Route } from 'react-router-dom';
 import Home from './pages/homeVisitor';
 import Signup from './pages/Signup';
 import Thanks from './pages/thanks';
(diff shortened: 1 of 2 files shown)


src/visitors/index.js
```diff
@@ -1,21 +1,23 @@
...
-  <Fragment>
+  <Switch>
     <Route exact path="/visitor/home" component={Home} />
     <Route path="/visitor/signup" component={Signup} />
     <Route exact path="/visitor/login" component={Login} />
     <Route exact path="/visitor/qrerror" component={QrError} />
     <Route exact path="/visitor/end" component={redirectAfterTimeout('/visitor/home', 5000)(Thanks)} />
     <Route exact path="/visitor/thankyou" component={redirectAfterTimeout('/visitor/home', 5000)(ThanksFeedback)} />
-  </Fragment>
+    <Route path="/visitor" component={NotFound} />
```
eliascodes: Here we specify `path`, but in the CB routes we don't. What's the difference, and which should we prefer?

astroash: Good spot, the path is already specified by the parent router so this should be removed.

astroash: fixed in amend

eliascodes:
Should I squash this one instead?

astroash:
You can't squash one commit.

Issue #652, opened 2019-01-24
🛀🏽 Suggested file (re)-structure


astroash:
Our current file structure does not lend itself to easily separating out pages into individual components. This is because any components that are pulled out of the `PageComponent` file end up in a components folder shared with all other non-page components. 

Suggestion:
- Each page is its own folder
- page components live in index.js
- pulled out 'page specific' components live in page folder
- `/components` renamed to `/sharedComponents`

edit:

On second thought, can be used **only** on pages that have separated out components and work with our current file structure.

eliascodes:
Yep, this sounds good to me.

PR #653, opened 2019-01-24
✨ Visitor sign in with name feature

### LINKED ISSUE #606: Address remaining comments in #600
#600 

Will be merged for convenience/expediency, but has issues that should still be addressed. See PR for them.

eliascodes:
Fixes #606 
Related to #536 

### Changes
* Adds sign-in with user-details step, starting with name, cascading through postcode, YOB, phone number and email, in that order (the order can be easily changed)

**Partnered PR**: TwinePlatform/twine-api#290

### Additional user flows to test
* [x] Sign in visitor with name (where name is unique)
* [x] Sign in visitor with name (where postcode is additionally required)
* [x] Sign in visitor with name (where YOB is additionally required)
* [x] Sign in visitor with name (where phone number is additionally required)
* [x] Sign in visitor with name (where email is additionally required)

More seed data has been added to the DB in the partnered PR in order to be able to exercise these cases. See the PR/Seed file for a description and for the visitor names/details to use.


astroash:
@eliascodes does this PR include changes outlined in #606 ?

astroash:
Initial feedback:
- feedback icons have disappeared on hosted app (not checked locally)
- I'm not keen on the second stage button change. [ sign in ] -> [ sign in with QR code ]. It feels jarring having the buttons stay in the same place and the text change. I actually clicked back and forth a number of times thinking this was a bug before realising what was happening. I think I prefered the previous version with the page split. What was the reasons for moving away from this?

astroash:
@eliascodes can the witch changes be pulled out to a separate PR? Squash & merge is being used so that we can revert features easily in the future. If we decide not to implement this feature we don’t want to loose these styling changes.

astroash:
Initial user flow feedback:

- 'Log a new visit' I think we should leave this as 'sign in'. It's more standard to see this on a visitor tracking dashboard, these users don't really care about the fact that their visit is _logged_
- signin options feel too cramped at the top of the page, can these be vertically centered? Will this impact the multiple steps?
screenshot 2019-01-30 at 11 44 51

src/cb_admin/pages/Activities.js
```diff
@@ -169,25 +171,35 @@ export default class ActivitiesPage extends React.Component {
...
-            id="cb-admin-activities-category"
-            label="Category"
-            name="category"
-            options={this.state.categories}
-            error={errors.activity}
-            required
-          />
-          <SubmitButton type="submit">ADD</SubmitButton>
-        </Form>
+        <Grid>
```
astroash: Can these `Grid` changes be separated out into a separate PR? I don't want them to be squashed into this PR incase this feature is never implemented.

src/tests/mock_instascan.js
```diff
@@ -0,0 +1,11 @@
+export default {
```
astroash: Can this also be separated out into a separate PR?

src/cb_admin/pages/Activities.js
eliascodes: From what I can see, the only changes that can be separated are those to `homeVisitor.js`. Just want to confirm that this is all you were expecting.

astroash: @eliascodes were the other changes from the styling PR?

src/tests/mock_instascan.js
eliascodes: See #662

astroash:
@eliascodes I've rebased `beta-testing` but the diff has not updates. Looks like we may need to open another PR again.

eliascodes:
It's because I hadn't rebased yet. It's done now, the diff dropped by ~100 lines.

src/visitors/components/ActivitiesGrid.js
```diff
@@ -0,0 +1,54 @@
...
+import { Row, Col } from 'react-flexbox-grid';
+import ActivitiesGridItem from './ActivitiesGridItem';
+import { pairs } from '../../util';
+
+
+const MarginSpacedCol = styled(Col)`
+  margin: 1em 0;
+`;
+
+const ActivitiesGrid = props => (
```
astroash: Can this be renamed `SignInActivitiesGrid`?

src/visitors/components/ActivitiesGridItem.js
```diff
@@ -0,0 +1,47 @@
...
+  height: 100%;
+  padding: 2em 0;
+`;
+
+const ActivitiesGridItem = props => (
+  <>
+    {
+      props.index % 2 === 0
+        ? (
+          <BtnOne name={props.name} onClick={e => props.onClick(e.target.name)}>
```
astroash: This layout is a bit difficult to read. Is it possible for the component to live on the same line as the ternary or is this an eslint thing?

src/visitors/components/FeedbackButtons.js
```diff
@@ -0,0 +1,49 @@
...
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { Row, Col } from 'react-flexbox-grid';
+import { PrimaryButtonNoFill } from '../../shared/components/form/base';
+import confused from '../../shared/assets/icons/faces/confused.svg';
+import happy from '../../shared/assets/icons/faces/happy.svg';
+import sad from '../../shared/assets/icons/faces/sad.svg';
+
+
+const FeedbackButton = styled(PrimaryButtonNoFill)`
```
astroash: Can this refactoring be pulled out into a separate PR?

src/visitors/components/QrScanner.js
```diff
@@ -0,0 +1,99 @@
...
+const instascanAvailable = () => {
+  try {
+    return Boolean(Instascan.Scanner);
+  } catch (error) {
+    return false;
+  }
+};
+
+
+export default class QrScanner extends React.Component {
```
astroash: Can this refactoring be pulled out into a separate PR?

src/visitors/pages/homeVisitor.js
```diff
@@ -44,12 +44,12 @@ export default class HomeVisitor extends Component {
         <Row around="xs">
           <Col xs={4}>
             <Link to="/visitor/login">
-              <ButtonLeft>Log new visit</ButtonLeft>
+              <ButtonLeft>Sign up</ButtonLeft>
```
astroash: up -> in

astroash:
@eliascodes what do you think about test user data being created for the various branches of this feature and added to stagings db?

src/visitors/components/ActivitiesGridItem.js
eliascodes: It's a linter thing, max line limit

eliascodes:
Is there anything else you want pulling out?

eliascodes:
> what do you think about test user data being created for the various branches of this feature and added to stagings db?

Can you explain this more?

astroash:
In the body of the PR you mention:

> The above user flows will probably need to be simulated by adding new visitors with the requisite conflicts/overlaps in the personal information (e.g. add two users with same name, YOB, null phone number, but different post codes in order to test the second user flow above).

It seems unuseful to do this manually & locally. If we create seed data for twine-api for these requirements they can be added to the staging or potentially live db so this feature can be easily demo'd.

astroash:
> Is there anything else you want pulling out?

Nothing that jumps out, but you're more familiar with the changes here. It's worth going through the PR and pulling anything out that is refactoring rather than new feature changes.

src/visitors/components/FeedbackButtons.js
eliascodes: See #663

src/visitors/components/QrScanner.js
eliascodes: See #664

src/visitors/pages/homeVisitor.js
eliascodes: Done in #663

astroash:
> The alternative is to set a fixed height of ~6em

@eliascodes is it possible to set a min height of ~6em and overflow? Then we have uniform for 1/2 line descriptions, but can accommodate for longer names too.

src/styles/index.css
```diff
@@ -8,3 +8,7 @@ body {
   font-size: 16px;
   overflow: hidden;
 }
+
+.full-height {
```
astroash: Why is this added to css not styled components?

src/visitors/components/SignInActivitiesGrid.js
```diff
@@ -0,0 +1,54 @@
...
+import SignInActivitiesGridItem from './SignInActivitiesGridItem';
+import { pairs } from '../../util';
+
+
+const MarginSpacedCol = styled(Col)`
+  margin: 1em 0;
+`;
+
+const SignInActivitiesGrid = props => (
+  <>
```
astroash: I've not seen this before, is it okay to give an unnamed element to jsx?

eliascodes: It's short-hand for `<Fragment>`

src/visitors/components/SignInForm.js
```diff
@@ -0,0 +1,172 @@
...
+
+export default class SignInForm extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      fields: SignInForm.Fields.slice(0, 1),
+      form: {},
+      errors: [],
+      uuid: btoa(new Date().toISOString()), // Used to prevent browser autocomplete
```
astroash: where is `btoa` coming from

eliascodes: It's a browser global: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa

src/visitors/components/SignInForm.js
```diff
@@ -0,0 +1,172 @@
...
+    );
+  }
+}
+
+SignInForm.propTypes = {
+  onSuccess: PropTypes.func.isRequired,
+  onFailure: PropTypes.func.isRequired,
+};
+
+SignInForm.MultipleUserError = class MultipleUserError extends Error {};
```
astroash: I'm unfamiliar with this syntax, what's going on here?

eliascodes: You mean extending `Error`? It's just a way to have more specific errors (so that you can tell them apart without coupling code to the specific error message)

src/visitors/pages/Login.js
```diff
@@ -0,0 +1,255 @@
...
+const BeatLoader = styled(Bl)`
+  width: 60px;
+  margin: 0 auto;
+  padding-top: 45vh;
+`;
+
+const capitaliseFirstName = name => name.split(' ')[0].replace(/\b\w/g, l => l.toUpperCase());
+const greetingWithOrgName = orgName => `Welcome${orgName ? ` to ${orgName}` : ''}!`;
+
+export default class Login extends Component {
```
astroash: Can this component be renamed to `SignIn`. `LogIn` is misleading to it's purpose.

src/visitors/pages/SignIn.js
```diff
@@ -0,0 +1,255 @@
+/*
+ *           ---- Scan QR code ---                       ---- Load activities ---- Register visit
+ *          /                     \                    /
+ * Login --                        -- Get visitor ID --
```
astroash: Login -> SignIn

src/visitors/components/SignInForm.js
astroash: And it's being stored on the `SignInForm` form `class`? I guess I am a little lost at the thought of a class being stored on another class (if that's what's happening)

src/visitors/components/SignInForm.js
```diff
@@ -0,0 +1,172 @@
...
+      errors: [],
+      uuid: btoa(new Date().toISOString()), // Used to prevent browser autocomplete
+    };
+  }
+
+  handleFormChange = (e) => {
+    this.setState(assocPath(['form', e.target.name.split('$')[0]], e.target.value));
+  }
+
+  handleFormSubmit = (e) => {
```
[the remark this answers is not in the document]
eliascodes: I'm trying to move towards components that are self-contained as far as reasonable. We have had some quite unwieldy and enormous "container" components, and I don't think that's the best way to do things. Also isolating all state to the page component even if it's only useful to one child is introducing an unnecessary coupling.

IMO the logic for determining
1. whether one or more matching users have been found
2. which fields to ask for next in the case of multiple users

should definitely not be the concern of the parent component. From the parent's perspective, it only needs/wants to know if there's been an error (and what kind), or if the sign in has been successful. The job of interpreting and processing API response data is not something that needs to live in the parent.

I think the idea of container/presentational components is fine, but there's nothing that says there can be only one container (at the top of the hierarchy). Containers are merely meant to decouple APIs from markup, and allow the use of prop-types to loudly indicate when API contracts change.

With that said, I think there are a couple further steps I could have taken here:
1. Either, pass in the specific API call as a prop from the parent (`Visitors.get`), then let the component deal with the rest.
2. Or, introduce a container `SignInFormContainer` that is responsible for the API call and passes data to a `SignInForm` child whose public API is closely defined via prop-types.

astroash:
@eliascodes can this feature use the file structure outlined in #652

eliascodes:
Should I convert the whole `visitor` directory or just that page?

astroash:
Just the feature/additions of this PR

astroash:
Multiple inputs rendered if you submit more than once:
sign_in_error_1

eliascodes:
Ooh good spot, are you double clicking?

astroash:
> Ooh good spot, are you double clicking?

Yes. If I am on a slow connection I can click 10 times and get 10 input boxes render.

astroash:
@eliascodes none of the search fields are using trim. I think that spaces should be trimmed from all values & removed from the middle of phone numbers.

eliascodes:
I don't think we're trimming spaces anywhere else. I agree we should be doing it, but I think this is an issue that should be fixed across the whole app, I don't think it's necessarily the best idea to implement it here first and then everywhere else.

astroash:
> I don't think it's necessarily the best idea to implement it here first and then everywhere else

Fair. This should probably be added to joi validation as a matter of urgency. Searching features are useless without this.

astroash:
User feedback:
- postcode is not displayed on the "Visitor Details" page. This makes it difficult to double check/update user details for the sign in with name feature

eliascodes:
> postcode is not displayed on the "Visitor Details" page.
> phone number is not being displayed on the "Visitor Details" page

These aren't specific issues with this feature, so should be done on staging. The beta branch can be rebased on-top of staging afterwards.

> reset button under the sign in button that clears any appended input boxes

Yeah, good shout, will add this.

astroash:
> These aren't specific issues with this feature, so should be done on staging.

Agreed. I'll make issues for them

astroash:
Did you intend the reset to be link text rather than a button?

screenshot 2019-02-11 at 13 25 02

eliascodes:
Yes, I wanted it to be visually distinct from the sign in button. The sign in button should be the visually dominant thing.

PR #655, opened 2019-01-28
✨ add Create Anon button for admin dashboard

### LINKED ISSUE #650: 👆🏽 Create Anonymous User page for admin section
- create button on dashboard for "Create Anonymous User"
- create form page for entering anonymous user data
  - name
  - age dropdown
  - gender dropdown
- submit button
- display QR code printout page

#536

astroash:
Fixes #650 
Related to #536 #652

#### Changes
* refactor printable qr code view to own component
- use new file structure outlined in 
- create page with multiple steps
- add link button to dashboard

User flows to test:
- go to admin dashboard
- create an anon user
- print anon user qr code

(following user tested needed due to refactoring)
- go to specific user
- print qr code
- sign up new user
- print qr code


src/cb_admin/pages/AnonUser/index.js
```diff
@@ -0,0 +1,97 @@
...
+        const getGenders = Visitors.genders();
+
+        return Promise.all([getCb, getGenders]);
+      })
+      .then(([{ data: { result: cbRes } }, { data: { result: gendersRes } }]) =>
+        this.setState({
+          cbOrgName: cbRes.name,
+          cbLogoUrl: cbRes.logoUrl,
+          organisationId: cbRes.id,
+          genders: [{ key: 0, value: '' }].concat(gendersRes.map(renameKeys({ id: 'key', name: 'value' }))),
```
eliascodes: This line of code appears around 5 times in the code base, probably time to think if there's a decent way to put that common code.

Looks like the underlying thing we're trying to do is transform an API payload (a list of objects) to a list of object values that can be used as options in a `select` element.

IMO this could live somewhere in `/api/...` or in `/shared/constants.js` since it has something to do with both things. Maybe worth considering if there's any other common/repeated transformations we're doing with API payloads, it could form the basis of an API transform/util module.

src/cb_admin/pages/AnonUser/index.js
```diff
@@ -0,0 +1,97 @@
...
+    switch (state.stage) { // eslint-disable-line default-case
+      case stage.ONE:
+        return (<CreateAnonUserForm
+          name={state.name}
+          handleChange={this.handleChange}
+          onSubmit={this.onSubmit}
+          errors={this.state.errors}
+          genders={this.state.genders}
+        />);
+      case stage.TWO:
```
eliascodes: I don't know if the concept of "stages" makes sense in this form. Are you anticipating more complex flows on this page? If not, a binary state is more simply represented with a boolean (`hasSubmitted` or something), rather than an enumeration. And it means you avoid a `switch` which has only two cases (aside: I don't think we should miss out `default` case for the same reason that we should always be very careful about missing out `else` statements after `if`s).

src/cb_admin/pages/AnonUser/CreateAnonUserForm.js
```diff
@@ -0,0 +1,78 @@
...
+`;
+
+const CenteredParagraph = styled(Paragraph)`
+  width: 90%;
+  font-size: 1.1em;
+  text-align: center;
+`;
+
+const CreateAnonUserForm = props => (
+  <FlexContainerCol>
```
eliascodes: If possible I think we should try to move away from our home-rolled layout system towards using `react-flexbox-grid` (I've been trying to do this in my recent PRs). I remember you asked for a quick run through about what I was doing, still happy to do that.

src/cb_admin/pages/AnonUser/PrintOption.js
```diff
@@ -0,0 +1,84 @@
...
+      <FlexContainerCol>
+        <NavHeader
+          leftTo="/cb/dashboard"
+          leftContent="Back to dashboard"
+          centerContent="Anonymous User QR code"
+        />
+        <CenteredParagraph>
+        Please print this page. If you loose the QR code it can be reprinted from the Visitor page.
+        </CenteredParagraph>
+        <FlexContainerRow>
```
eliascodes: Same layout comment as above.

src/cb_admin/pages/AnonUser/index.js
astroash: @eliascodes can you make an issue for this so it can be resolved separately?

src/cb_admin/pages/AnonUser/index.js
astroash: Yeah, this was a stylistic choice. We should research best practices for displaying react content based on state to see if we should adopt convention moving forward.

Let's break this into an issue too 🙂

astroash:
@eliascodes let's hold off merging until all `staging` PRs are in

src/cb_admin/pages/AnonUser/index.js
eliascodes: See #659

src/cb_admin/pages/AnonUser/index.js
eliascodes: See #660

PR #656, opened 2019-01-28
🚀 DEPLOYMENT 28-01-19

astroash:
Standard User Flows:
- [x]         signup visitor
- [x]         log visit
- [x]         export logs

Additional things/user flows to test:

- [x] navigation header on all pages
- [x] sign up with only email
- [x] sign up with only phone number

Partnered PR: TwinePlatform/twine-api#320

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -2,7 +2,7 @@
  * Twine API interface
  */
 import _axios, { create } from 'axios';
-import { pathOr, equals, evolve, map } from 'ramda';
+import { pathOr, equals, evolve, map, filter } from 'ramda';
 import qs from 'qs';
 import { BirthYear } from '../shared/constants';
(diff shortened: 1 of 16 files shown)


eliascodes:
Signing up without a phone number seems to work.

When trying to sign up as a visitor with phone number only, there's a 500 response, which results from this error:

screenshot 2019-01-28 at 11 14 08

astroash:
@eliascodes what org are you signed in as? Can you run this sql command on your db 

```
SELECT * from user_account 
  INNER JOIN user_account_access_role 
    ON user_account_access_role.user_account_id = user_account.user_account_id 
  WHERE user_account.email LIKE '%anon%' 
  AND user_account_access_role.organisation_id = [id]
```

PR #657, opened 2019-01-28
💄 Styling fixes & prop bug fix

eliascodes:
This is mainly to get away from the silly 90% thing we were
doing everywhere. Using 100% width and setting explicit margin
and padding is better, instead of having layouts implicitly depend
on elements not using the full width of their container.

* Make LabelledInput and LabelledSelect 100% width
* Reduce padding on LabelledSelect to improve alignment w LabelledInput
* Update several pages w/ Button width 90% -> 100%
* Add margin/padding to places that were relying on this 90% width thing

In addition, fix a bug with the newly introduced `NavHeader` component; `Fragment` elements are not allowed to have click handlers.

### DIFF
--- a/src/cb_admin/components/DetailsTable.js
+++ b/src/cb_admin/components/DetailsTable.js
@@ -9,9 +9,10 @@ const Table = styled.table`
   background-color: ${colors.highlight_secondary}; /* Fallback */
   background: linear-gradient(0, ${rgba(colors.highlight_secondary, 0.75)} 0%, ${colors.highlight_secondary} 100%);
   color: ${colors.white};
-  width: 90%;
-  height: 6em;
-  padding: 2em;
+  height: 98%;
+  width: 100%;
+  padding: 1.5em 2em 1.5em 1em;
+  word-break: break-word;
 `;
 const TableCaption = styled.caption`
   display: none;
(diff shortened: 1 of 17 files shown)


astroash:
@eliascodes I feel like now we're missing padding on a number of the pages

eg.
screenshot 2019-01-29 at 21 40 12

eliascodes:
Where is there missing padding on that page?

astroash:
@eliascodes re conversation, see alignment changes on settings page:

Before:
screenshot 2019-01-30 at 10 32 53

After:
screenshot 2019-01-30 at 10 32 50

The changes in input/button width has changed layout throughout the app. IMO 90% width should be replaced with margin.

astroash:
Input boxes are looking a bit janky on the settings page. I think they should either be centred or the same width as the box above.

screenshot 2019-01-30 at 14 18 46

eliascodes:
Hmm, missed this because I was only looking at it on iPad resolution.

astroash:
Can we add a little padding to the signup form 🙏🏽. It looks so much nicer!

Without padding:
screenshot 2019-01-30 at 16 39 51

With padding:
screenshot 2019-01-30 at 16 39 58

Also, happy for this to be solved in another PR, but as there's no set height on the `img` element there's a very obvious page rejig on response. This was there before, but the `img` was smaller so it was less obvious:
settings page

eliascodes:
> Can we add a little padding to the signup form 🙏🏽. It looks so much nicer!

I can barely tell the difference. Is this blocking?

astroash:
No...

PR #658, opened 2019-01-29
🐛 fix signup bug

astroash:
discovered in review #656

#### Changes
- do not send empty values in signup flow
- render correct error messages

User flow:
- sign up with only email
- signup with only phone number

Partnered PR: TwinePlatform/twine-api#326

### DIFF
--- a/src/visitors/components/signup_form.js
+++ b/src/visitors/components/signup_form.js
@@ -73,7 +73,7 @@ const signupForm = (props) => {
                 label="Full Name"
                 name={`fullname$${props.uuid}`}
                 type="text"
-                error={props.errors.formSender && VISITOR_NAME_INVALID}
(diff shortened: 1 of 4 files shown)


src/visitors/components/signup_form.js
```diff
@@ -73,37 +73,37 @@ const signupForm = (props) => {
...
+                error={props.errors.name && VISITOR_NAME_INVALID}
                 required
               />
               <LabelledInput
                 id="visitor-signup-email"
                 label="Email Address"
                 name={`email$${props.uuid}`}
                 type="email"
-                error={props.errors.formEmail}
+                error={props.errors.email}
```
eliascodes: `formEmail` is used in the `409` error case in the `pages/Signup.js` component, so needs to be updated to work with this change

PR #663, opened 2019-01-31
♻️ extract feedback buttons

eliascodes:
also styling and copy improvements

Related to #653 

#### Changes
* Extract feedback buttons component
* Use flexbox to style `homeVisitor`
* Update copy on buttons on `homeVisitor`


astroash:
The layout of the main visitor page has significantly changed.

Before:
live-visitor-page

After:
pr-visitor-signin-page

eliascodes:
Should be much closer to the original now, but the feedback smileys have been spaced a little further apart (may prevent mis-clicks).

PR #693, opened 2019-02-14
⬆️ Update dependencies

eliascodes:
Except for those likely to have breaking changes

* Major bumps: jest-dom
* Minor bumps: eslint-plugins, moment, polished, prop-types, react, ramda, qs, react-scripts, react-spinners, styled-components, react-testing-library, axios-mock-adapter

**Changes:**
eslint-plugin-react 
- no breaking changes, see changelog https://github.com/yannickcr/eslint-plugin-react/blob/master/CHANGELOG.md

moment 
- no breaking changes, see changelog https://github.com/moment/moment/blob/develop/CHANGELOG.md

polished
- no breaking changes, see semantic versioning https://github.com/styled-components/polished/releases?after=v2.3.1

qs
- no breaking changes, see changelog https://github.com/ljharb/qs/blob/master/CHANGELOG.md

ramda
- a number of deprecated methods, all unused in codebase
- see changes here https://github.com/ramda/ramda/issues/2702
- note evolve now works with arrays ✨

react
- no breaking changes, see changelog here https://github.com/facebook/react/blob/master/CHANGELOG.md
- note hooks now stable ✨

react-scripts
- no breaking changes, see changelog https://github.com/facebook/create-react-app/blob/master/CHANGELOG.md

styled-componenets
- no breaking changes, see changelog https://github.com/styled-components/styled-components/blob/master/CHANGELOG.md

axios-mock-adapter
- no breaking changes, see semantic versioning https://github.com/ctimmerm/axios-mock-adapter/releases

jest-dom
- breaking changes:
  - toHaveTextContent to be case-sensitive
- see semantic versioning https://github.com/gnapse/jest-dom/releases

react-testing-library
- no breaking changes, see semantic versioning https://github.com/kentcdodds/react-testing-library/releases

PR #696, opened 2019-02-19
🚀 DEPLOYMENT 19-02-19

astroash:
Changes
- 📝 add docs - initial pass of user testing instructions (#681)
- ⬆️ Update dependencies to minor releases (#693)

Partnered PR: https://github.com/TwinePlatform/twine-api/pull/346

Manual testing:

- [x] `CB_ADMIN` can successfully delete an activity
- [x] `CB_ADMIN` can update a `VISITOR`s name to different casing eg ASH -> ash
- [x] as this PR includes dependency changes common user flows should also be tested for unanticipated bugs


mandsmandsmands:
I've just tested (IE and Chrome)
**CB_ADMIN can successfully delete an activity**
on https://twine-visitor-staging.herokuapp.com and I am still getting  - 500 internal server error

astroash:
@mandsmandsmands 

>on [twine-visitor-staging.herokuapp.com](https://twine-visitor-staging.herokuapp.com) and I am still getting - 500 internal server error

Thanks for highlighting this, this should not be the case.

> * `CB_ADMIN` can update a `VISITOR`s name to different casing

Name fields should be able to change casing or input a new name. Previously we had a 🐛 where case changes would not work

>     * VISITOR - I do not have QR code for this env so need some help with this.

You can either:
- create a new QR code by visitor sign up. For demoing I often create fake email accounts and take a picture of the QR code on my phone
- Go to the Visitors page on the admin dashboard and select (click on) a visitor. This is take you to their details page where you can take a photo of the QR code

eliascodes:
@mandsmandsmands thanks for that!

On your FAIL, it was indeed designed to allow the name to be totally changed, so that's fine (as far as we're concerned). The only thing that was added in this PR was a bug fix allowing casing-only changes to be recognised.

For example, before this, the change `Seldwyn Morgan` -> `seldwyn morgan` wouldn't work (it would just give you the original back again). But `Seldwyn Morgan` to `seldwyn m` would work fine.

> VISITOR - I do not have QR code for this env so need some help with this.

The way we typically do this is one of two ways:
1. Sign up as a new visitor, and then take a picture of the QR code when it's shown on screen
2.  Go to the admin section, to the list of visitors, chose one and take a picture of their QR code when it's shown on their profile view.

eliascodes:
> Thanks for highlighting this, this should not be the case.

This seems to be because the new version was not deployed. I think auto-deploy doesn't go through if any of the checks fail (including coverage). So because coverage dropped (even by <1%) it didn't deploy.

I've manually deployed to staging now though, should be fixed and working.

PR #703, opened 2019-03-06
🚀 DEPLOYMENT 06-03-19

astroash:
**Changes**

- Create a new simpler/better custom checkbox
- Use it on the visitor details page to allow contact consent to be changed

**Manual testing:**

- [x] Change contact consent of a visitor and press save
- [x] if I add an activity with spaces at the beginning of the name they are removed when added

**Partnered PR:** https://github.com/TwinePlatform/twine-api/pull/367

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -77,7 +77,16 @@ export const Visitors = {
       }),
     ),
 
-  update: ({ id, name, gender, birthYear, email, phoneNumber } = {}) =>
+  update: ({
+    id,
+    name,
+    gender,
+    birthYear,
+    email,
+    phoneNumber,
+    isEmailConsentGranted,
+    isSMSConsentGranted,
+  } = {}) =>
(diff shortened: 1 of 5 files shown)


mandsmandsmands:
**Change contact consent of a visitor and press save**
Tests:
All 4 cases originally False / False.
Changed to state, then changed again.
Email True / SMS True > Email False / SMS False
Email True / SMS False > Email False / SMS True 
Email False / SMS True > Email True / SMS False 
Email False / SMS False  > Email True / SMS True

New User created: 
Email False / SMS False  > Email True / SMS True

mandsmandsmands:
**if I add an activity with spaces at the beginning of the name they are removed when added**
PASS
1. Add activity with space, check on Activities List
2. Check on Visitor sign-in, select Activity 
3. Check Visits Data > Activity filter drop-down, pie chart and table

PR #704, opened 2019-03-21
✨ add confirm role page

astroash:
Related to TwinePlatform/hq#57

#### Changes:
- accept values from querystring and url
- attempt to create account
- display message if visitor account is created
- display error message if there was a problem
- redirect to pw reset if volunteer account

Partnered PR: https://github.com/TwinePlatform/twine-api/pull/371

#### Manual Testing:
- visitor second role
  - sign up to visitor app with email used as a volunteer
  - respond to confirmation email
  - webpage will tell you its a success
- volunteer second role
  - sign up to volunteer app with email used as a visitor
  - respond to confirmation email
  - webpage will redirect you to password reset to set your password
- error 
  - go to page with a used cookie will give you an error message
  - go to page with bad/edited query string will give you an error message

#### Notes:
- no testing in favour of adding once monorepo is set up

### DIFF
--- a/src/confirm_role/pages/ConfirmRole.js
+++ b/src/confirm_role/pages/ConfirmRole.js
@@ -47,24 +47,15 @@ export default class ConfirmRole extends Component {
     const { userId, organisationId, role } = parse(search.replace('?', ''));
     Visitors.addRole({ token, userId, organisationId, role: role && role.toUpperCase() })
       .then((r) => {
-        console.log({ r });
-
         const res = ResponseUtils.getResponse(r);
-        console.log({ res });
-
         return res.token
           ? this.setState({ status: status.REDIRECT, user: res })
           : this.setState({ status: status.SUCCESS });
       })
-      .catch((e) => {
-        console.log(e);
-
-        this.setState({
-          status: status.FAILURE,
-          errors: ErrorUtils.getErrorMessage(e),
-        })
-        ;
-      });
+      .catch(e => this.setState({
+        status: status.FAILURE,
+        errors: ErrorUtils.getErrorMessage(e),
+      }));
   }
 
   render() {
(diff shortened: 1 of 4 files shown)


src/confirm_role/pages/ConfirmRole.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  componentDidMount() {
+    const { props: { location: { search }, match: { params: { token } } } } = this;
+    const { userId, organisationId, role } = parse(search.replace('?', ''));
+    Visitors.addRole({ token, userId, organisationId, role: role && role.toUpperCase() })
+      .then((r) => {
+        const res = ResponseUtils.getResponse(r);
+        return res.token
+          ? this.setState({ status: status.REDIRECT, user: res })
```
eliascodes: Is this case possible? Looking at [`/register/confirm`](https://github.com/TwinePlatform/twine-api/pull/371/files#diff-3f19b613b2c353cbf665d7d878b06b61), it looks like you only get a token if you are registering a VOLUNTEER role to a pre-existing VISITOR account, which can't happen on the visitor app, no?

astroash: Both emails from the visitor app and volunteer app redirect to this page.

src/confirm_role/pages/ConfirmRole.js
```diff
@@ -0,0 +1,104 @@
...
+      })
+      .catch(e => this.setState({
+        status: status.FAILURE,
+        errors: ErrorUtils.getErrorMessage(e),
+      }));
+  }
+
+  render() {
+    return (
+      <FlexContainerCol justify="flex-start">
```
eliascodes: We should stick to `react-flexbox-grid` instead of the old components.

Issue #707, opened 2019-04-08
QA: Confirm Role Feature


eliascodes:
Related to #704 and TwinePlatform/twine-api#371
### `visitor-app`
* Case 1: **SUCCESS** - `VOLUNTEER` user adds a second role (`VISITOR`)
  - [x] step 1
  * populate DB with `VOLUNTEER` user with e-mail that you have access to
  * sign up to visitor app with email used as a `VOLUNTEER`
  * ~~respond to confirmation email~~
  * ~~webpage will tell you its a success~~
  - [x] step 2
  * go to email and click on link (domain will need to be replaced with `localhost:3000`)
  * confirm your email creation
  * check you have an email with your QR code
* [x] Case 2: **FAILURE** - `VOLUNTEER_ADMIN` user adds a second role (`VISITOR`)
  * populate DB with `VOLUNTEER_ADMIN` user with e-mail that you have access to
  * sign up to visitor app with email used as a `VOLUNTEER`
  * webpage will tell you its a failure (see new PR)
* [x] Case 3: **FAILURE** - `VOLUNTEER` user from another CB adds a second role (`VISITOR`)
  * populate DB with `VOLUNTEER` user at another CB with e-mail that you have access to
  * sign up to visitor app with email used as a `VOLUNTEER`
  * webpage will tell you its a failure

### `volunteer-app`
* [x] Case 1: **SUCCESS** - `VISITOR` user adds a second role (`VOLUNTEER`)
  * reset the DB, insert demo data
  * populate DB with `VISITOR` user with e-mail that you have access to
  * sign up to volunteer app with email used as a `VISITOR`
  * respond to confirmation email
  * webpage will redirect you to password reset to set your password
  * reset password
  * log in to volunteer app
* [x] Case 2: **FAILURE** - `VISITOR` user adds a second role (`VOLUNTEER_ADMIN`)
  * sign up to volunteer app with email used as a `VISITOR` & a `VOLUNTEER_ADMIN` code
  * webpage will tell you its a failure
* [x] Case 3: **FAILURE** - `VISITOR` user from another CB adds a second role (`VOLUNTEER`)
  * populate DB with `VISITOR` user at another CB with e-mail that you have access to
  * sign up to volunteer app with email used as a `VISITOR`
  * webpage will tell you its a failure

### security
* [x] Case 1: token already used
  * go to confirm page with a used token will give you an error message (URL path of the form `/confirm-role/USEDTOKENGOESHERE?role=...`
[...]

PR #709, opened 2019-04-10
🐛 fix confirm role redirect

eliasmalik:
Addresses bug found in #707


eliascodes:
🚨🚨🚨 Self-merging b/c of urgency and @astroash is off at the mo 🚨🚨🚨

